PR #49, opened 2014-11-04
Add GridEngine multiplier capability to parser

tofu-rocketry:
Replacement pull request for #36.

The GridEngine scheduler does not include any built-in solution to normalize the cputime/walltime numbers in a heterogeneous cluster. This change extends GridEngine's APEL parser to normalize these values according to the nodes' definition.

For each accounting record, cputime/walltime are computed based on the 'cputmult' and 'wallmult' (a la PBS) complex values set in the node's definition. If no multiplier is found, it defaults to 1.

### DIFF
--- a/apel/parsers/sge.py
+++ b/apel/parsers/sge.py
@@ -66,6 +66,8 @@ class SGEParser(Parser):
             if p.returncode != 0:
                 raise MultiplierError(err)
         except (OSError, MultiplierError):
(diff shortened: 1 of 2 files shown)


tofu-rocketry:
Hooray! There's no other objections from the batch working group. :tada:

tofu-rocketry:
I've added a warning at L69.

PR #52, opened 2014-11-06
Add method stubs to base ApelDb class

tofu-rocketry:
These changes are mainly tidy-up as all that's being changed is that the class that invokes the actual database backend classes is being given method stubs that have the right signature (arguments). They all now raise `NotImplementedError` if they're called directly as the backend classes should override these.

(Also improves code health.)

It's not strictly an "abstract base class", though it's kind of acting like one.


tofu-rocketry:
Deleting from central repository. I'll still have the branch in my fork.

PR #63, opened 2015-01-16
Keep DNs if no valid response from GOCDB

### LINKED ISSUE #60: Keep existing dns file when GOCDB query fails
When there is an error querying the GOCDB, the `retrieve_dns.py` program creates a `dns` file with only the hard-coded DNs in it. Instead, it should leave the existing 'dns' file as it is. This way, the system will continue to operate with the DNs from the last successful query.

tofu-rocketry:
Resolves #60.

**Main changes:**
- Change `retrieve_dns` to keep the previous DN list if there isn't a valid
  response from GOCDB. This is so that messages aren't rejected just
  because there was a small period where GOCDB is unreachable.
- Add config option to alter the expiry time of the DN list when the above
  situation occurs.
- Add lots of test cases for `retrieve_dns`. :triumph:

**Minor changes:**
- Change `retrieve_dns` to only print a debug message when it encounters a
  comment line starting with "#" as there used to always be a few that
  would generate warnings unnecessarily.
- Reorder `sys.exit` and `log.info` in `retrieve_dns` so that the logging call
  is first otherwise it is never reached.

### DIFF
--- a/bin/retrieve_dns.py
+++ b/bin/retrieve_dns.py
@@ -31,6 +31,7 @@ import ConfigParser
 import logging.config
 import os
 import sys
+import time
 import urllib
 import xml.dom.minidom
 import xml.parsers.expat
@@ -46,6 +47,7 @@ class Configuration(object):
         self.banned_dns = None
         self.dn_file = None
         self.proxy = None
+        self.expire_hours = None
 
 def get_config(config_file):
     """Using the config file location, get a config object."""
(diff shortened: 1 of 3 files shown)


stuartpullinger:
Really good to see increases in the code coverage. Well done.

tofu-rocketry:
The reduction in repository health seems to be an artefact of Landscape adding cyclomatic complexity checks as they've popped up on functions that haven't been changed.

PR #64, opened 2015-01-29
Add method stubs to base ApelDb class

tofu-rocketry:
This is a new and improved version of #52 but this time just using `pass` instead of `raise NotImplementedError` so that it doesn't interfere with the Oracle back-end. Changes include:

- Add method stubs to base ApelDb class to show what should be implemented
  by database backends.
- Change method signature.


tofu-rocketry:
Should fix this properly using `NotImplemented` error and `pass` in the Oracle child class if necessary

PR #65, opened 2015-02-02
Add logging to star_parser

### LINKED ISSUE #62: Add logging to STAR parser
There is a `print` call in `star_parser.py` [here](https://github.com/apel/apel/blob/dev/apel/db/loader/star_parser.py#L123) that should be a logging call, though logging needs to be added to it in general. (c.f. `car_parser.py`, `aur_parser.py`)

The output of `print` can be seen in the output of the CI.

tofu-rocketry:
Resolves #62.

Simple fix for missing logging. Follows same pattern as for the [car_parser](https://github.com/apel/apel/blob/dev/apel/db/loader/car_parser.py).

### DIFF
--- a/apel/db/loader/star_parser.py
+++ b/apel/db/loader/star_parser.py
@@ -17,10 +17,16 @@
 '''
 
 
+import logging
+
 from apel.db.records.storage import StorageRecord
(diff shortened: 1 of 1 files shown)


stuartpullinger:
Great. Good spotting. Please merge.

PR #66, opened 2015-02-05
Univa Grid Engine timestamp option

### LINKED ISSUE #47: Allow for new version of Grid Engine reporting in milliseconds
As reported [on GGUS](https://ggus.eu/index.php?mode=ticket_info&ticket_id=109286).

Univa Grid Engine 8.2.0 now reports time differently (in a MINOR version upgrade :frowning: ). In the accounting log seconds are replaced with milliseconds in certain fields.

We need to decide how to handle this change (config, different parser?) and implement.

tofu-rocketry:
Resolves #47.

- Add option to parser config that defines whether Grid Engine is using
  millisecond timestamps as Univa Grid Engine is doing this from version
  8.2.0. Will default to not converting the timestamp if the option is
  absent.
- Add unit test for this new timestamp format and unit test for changes to `parser.py`.
- Move call to datetime function in `apel/db/records/record.py` to its own `try...except` block as it needs
  different handling to the `int` call when raising a `ValueError`. In the former
  case, it is likely to be a timestamp out of range (possibly caused by the
  Grid Engine oddness) and having it masked by the error message of the `int` call was confusing.
- Make `test_record.py` directly runnable.

### DIFF
--- a/apel/db/records/record.py
+++ b/apel/db/records/record.py
@@ -158,7 +158,6 @@ class Record(object):
                 # We accept ints or floats as seconds since the epoch
                 try:
                     value = int(value)
-                    return datetime.utcfromtimestamp(value)
                 except ValueError:
                     # Not a datetime or an int, so it has to be a string representation.
                     # We get ISO format dates when parsing CAR or StAR.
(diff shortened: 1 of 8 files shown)


stuartpullinger:
Excellent. Please merge.

PR #69, opened 2015-02-13
Log entries for saved and rejected messages

### LINKED ISSUE #67: Add message location to log output
This has already been done for partly in d8f5d60237d454800599502adbf66cd406996a40.

The message "name" (i.e. directory) should be printed (as well as the ID from the originating queue) when the message is added to the accept or reject queues in [`loader.py`](https://github.com/apel/apel/blob/dev/apel/db/loader/loader.py#L152).

tofu-rocketry:
Resolves #67.

- Added a bit more info to the log output when messages are saved or rejected so that the queue location is added to the log.
- Added the first unittests for `loader.py`.
- Also added a missing `os.remove` to a different test.

### DIFF
--- a/apel/db/loader/loader.py
+++ b/apel/db/loader/loader.py
@@ -149,19 +149,21 @@ class Loader(object):
                 self.load_msg(msg_text, signer)
                 
                 if self._save_msgs:
-                    self._acceptq.add({"body": msg_text,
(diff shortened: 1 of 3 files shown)


stuartpullinger:
Great. Please merge.

PR #72, opened 2015-03-04
Fix to migrate_apel.py date calculation 

tofu-rocketry:
- Change calculations for target month and year for setting the starting
  point of the migration as the existing calculations didn't really work
  over year boundaries.


stuartpullinger:
OK. It works for me. Please merge.

PR #73, opened 2015-03-04
Change default for 'parallel' to true

### LINKED ISSUE #68: Turn parallel job reporting on by default
We need to check the impact on the parsers. Then change the default config option to 'true' - which will only affect new installations.

tofu-rocketry:
Resolves #68.

- Change default for 'parallel' to true in parser config so that new
  installations will default to parsing MPI data.


stuartpullinger:
Great. Please merge.

PR #76, opened 2015-06-05
Normalised summaries aggregation

tofu-rocketry:
As the `VNormalisedSuperSummaries` view does not use the `ServiceLevelType` and `ServiceLevel` columns for grouping, which the primary key of the underlying `HybridSuperSummaries` table does, certain columns do not have the correct values. This can be see when comparing to `VSuperSummaries`. This pull request fixes those columns so that they aggregate data correctly, as follows:

- Take `SUM` of `NumberOfJobs`.
- Take `MIN` of `EarliestEndTime`.
- Take `MAX` of `LatestEndTime`.


stuartpullinger:
Great. Looks good. Please merge.

PR #84, opened 2015-09-16
ARC Parser

tofu-rocketry:
These changes add a basic ARC parser to APEL.

Note that ARC produces one accounting record per file, and that they contain enough information to produce a job record directly, so this influences where the ARC parser can fit in with our existing code without too many changes.

You may want to use [this link](https://github.com/apel/apel/compare/dev...tofu-rocketry:arc-parser?diff=unified&name=arc-parser&w=1) to view the diff with whitespace changes ignored. (Mainly for `bin/parser.py` which has a lot of the extant code indented into an `else:` block.)

### DIFF
--- a/apel/common/datetime_utils.py
+++ b/apel/common/datetime_utils.py
@@ -47,10 +47,16 @@ def parse_timestamp(datetime_string):
     '''
     Parse timestamp encoded as a string in various forms.  Return
     a TZ-unaware datetime object, which is in UTC.
-    
+
     If timezone information is not present in the string, it
-    assumes that the timezone is UTC.  
+    assumes that the timezone is UTC.
     '''
+
+    # If format is YYYYMMDDhhmmssZ (no T) add time designator so it is strictly
+    # compliant with ISO 8601. (Some versions of iso8601 module are fussy.)
+    if re.match(r'[0-9]{14}Z', datetime_string) is not None:
+        datetime_string = datetime_string[:8] + 'T' + datetime_string[8:]
+
     dt = iso8601.parse_date(datetime_string)
     utcdt = dt.astimezone(iso8601.iso8601.UTC)
     # internal representation of datetimes is UTC without timezones
--- a/apel/parsers/arc.py
+++ b/apel/parsers/arc.py
@@ -3,7 +3,7 @@ from apel.db.records.job import JobRecord
 from apel.parsers import Parser
 
 
-class ArcParser(Parser):
+class ARCParser(Parser):
     """
     Parser for ARC accounting records.
 
@@ -12,7 +12,10 @@ class ArcParser(Parser):
     """
     def parse_arc_file(self, arcfile):
         """
-        Takes an open ARC accounting file object and returns a JobRecord.
+        Parse an ARC accounting file.
+
+        Takes an open ARC accounting file object and returns a JobRecord and the
+        number of lines in it.
         """
         arcjob = {}
         lines = 0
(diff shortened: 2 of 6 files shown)


apel/common/datetime_utils.py
```diff
@@ -47,10 +47,16 @@ def parse_timestamp(datetime_string):
...
-    
+
     If timezone information is not present in the string, it
-    assumes that the timezone is UTC.  
+    assumes that the timezone is UTC.
     '''
+
+    # If format is YYYYMMDDhhmmssZ (no T) add time designator so it is strictly
+    # compliant with ISO 8601. (Some versions of iso8601 module are fussy.)
+    if re.match(r'[0-9]{14}Z', datetime_string) is not None:
```
kjopek: Shall we use regexp like:
``r^[0-9]{14}Z`` instead of the one above? It currently accepts every text containing YYYYMMDDhhmmssZ but it should accept only one starting with date.

apel/parsers/arc.py
```diff
@@ -0,0 +1,72 @@
...
+        arcjob = {}
+        lines = 0
+        for line in arcfile:
+            # Remove any leading or trailing whitespace including newlines.
+            line = line.strip()
+            lines += 1
+            key, value = line.split('=', 1)
+            arcjob[key] = value
+
+        if lines == 0 or 'status' not in arcjob or arcjob['status'] != "completed":
```
kjopek: This condition can be rewritten as:
``if lines == 0 or arcjob.get('status') != "completed":``

apel/parsers/arc.py
```diff
@@ -0,0 +1,72 @@
...
+            # Empty file, incomplete job, or failed job (respectively)
+            return None, lines
+
+        arcjob['accounting_options'] = self.parse_accounting_options(arcjob['accounting_options'])
+
+        apeljob = {}
```
kjopek: This could be rewritten as:
``apeljob = {
   'Site': arcjob['accounting_options']['gocdb_name'],
   'SubmitHost': arcjob['headnode']
``
and so on. There is a lot of repetitions of word 'apeljob'.

test/test_arc.py
```diff
@@ -0,0 +1,112 @@
...
+        incomplete_job = StringIO.StringIO(incomplete)
+
+        record, lines = self.parser.parse_arc_file(incomplete_job)
+        self.assertEqual(record, None)
+        self.assertNotEqual(lines, 0)
+
+        incomplete_job.close()
+
+
+complete = r"""loggerurl=APEL:http://mq.cro-ngi.hr:6162
```
kjopek: Why this string starts with r? It does not look like regular expression.

apel/parsers/arc.py
```diff
@@ -0,0 +1,72 @@
...
+        The ARC to APEL mapping used can be found in the JURA souce code here:
+        http://svn.nordugrid.org/trac/nordugrid/browser/arc1/trunk/src/services/a-rex/jura/arexlog2ur.txt
+        """
+        arcjob = {}
+        lines = 0
+        for line in arcfile:
+            # Remove any leading or trailing whitespace including newlines.
+            line = line.strip()
+            lines += 1
+            key, value = line.split('=', 1)
```
gregcorbett: Would be useful to explain why we only split on the first `=`.

test/test_arc.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+        invalid_file = StringIO.StringIO("valid=line\ninvalid line")
+
+        self.assertRaises(ValueError, self.parser.parse_arc_file, invalid_file)
+
+        invalid_file.close()
+
+    def test_parse_incomplete_job(self):
+        """
+        Test that the parser returns None and >0 for an incomplete job.
```
gregcorbett: Returns `>0`? Not just `0`?

tofu-rocketry: Ah, it's not that this function returns 0, but it's checking that another function returns `None` for one value of a tuple and a value greater than 0 for the other. That could be clearer.

gregcorbett (changes requested):
See comments

tofu-rocketry:
No need for this any more.

PR #88, opened 2015-12-11
Correct path to dirq Queue class

tofu-rocketry:
Change path to Queue class to explicitly go via its parent module as dirq version1.7 has removed the import of Queue in the `__init__.py`.


tofu-rocketry:
Increase in coverage is from Coveralls correcting the way it counts lines.

PR #89, opened 2015-12-11
Add support for Torque 5.1.2 time format

### LINKED ISSUE #87: Support new Torque 5.1.2 time format
[GGUS ticket.](https://ggus.eu/index.php?mode=ticket_info&ticket_id=118147)

Torque 5.1.2 (a patch release) has an undocumented backwards incompatible change. :sob:

Time is now reported in seconds rather than `hh:mm:ss`. Add support for this, while maintaining our own backwards compatibility. :innocent:

- [x] Check that it is now actually seconds.
- [x] Issue patched PBS parser pre-holidays.

tofu-rocketry:
Resolves #87.

- Add check for `wallduration` and `cput` being in seconds (though it just checks the latter) rather than
  `hh:mm:ss` format as Torque 5.1.2 has changed to using the former. The parser then chooses what method to use to parse the duration fields meaning the change is backwards compatible.
- Add more test cases.

### DIFF
--- a/apel/parsers/pbs.py
+++ b/apel/parsers/pbs.py
@@ -60,13 +60,20 @@ class PBSParser(Parser):
         else:
             nodes, cores = 0, 0
 
+        # Torque 5.1.2 uses seconds rather than hh:mm:ss for cput and walltime
+        # so check for that here.
+        if ':' not in data['resources_used.cput']:
+            time_function = lambda y: y
+        else:
+            time_function = parse_time
+
         # map each field to functions which will extract them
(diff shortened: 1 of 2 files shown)


tofu-rocketry:
Decided to add some more tests while I was in there are go for full coverage of that module.

tofu-rocketry:
Landscape doesn't seem to be processing code quality checks at the moment.

stuartpullinger:
Great. Things I like:

- Adding extra test cases
- Comments explaining code
- You are checking both cput and walltime to make the code more watertight.

Please proceed to merge.

PR #92, opened 2016-01-11
XML with no records crash fix

### LINKED ISSUE #85: Empty storage record crashes loader with wrong message
https://github.com/apel/apel/blob/dev/apel/db/loader/star_parser.py#L54

When an empty, but valid xml, storage record is loaded an exception is raised. The exception is not caught so it crashes the loader. We should move the empty message to the reject queue and continue.

tofu-rocketry:
Resolves #85.

- Change reference to CAR records to StAR records for exception message in
  StAR record parser-loader.
- Change exception from plain `Exception` to `XMLParserException` for StAR and
  AUR parser-loaders (to make sure it's consistent) as the plain `Exception` would not be caught by the
  loader. CAR parser already uses the [right exception](https://github.com/apel/apel/blob/dev/apel/db/loader/car_parser.py#L48).
- Add corresponding unit test for StAR parser-loader.

### DIFF
--- a/apel/db/loader/aur_parser.py
+++ b/apel/db/loader/aur_parser.py
@@ -17,7 +17,7 @@
 '''
 
 from apel.common import iso2seconds, parse_timestamp
-from xml_parser import XMLParser
+from xml_parser import XMLParser, XMLParserException
 from apel.db.records.normalised_summary import NormalisedSummaryRecord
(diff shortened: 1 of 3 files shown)


stuartpullinger:
Looks good. Please merge.

PR #94, opened 2016-02-09
Apel setup script

### LINKED ISSUE #86: Add setup.py script
We should add a `setup.py` script to allow our packages to be installed on any OS that has Python installed. It would also possibly help with automated testing.

https://docs.python.org/2.6/install/
https://docs.python.org/2.6/distutils/setupscript.html

Same issue for SSM: apel/SSM#31

gregcorbett:
resolves #86 

a python setup script, usage: python setup.py install

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,72 @@
+"""A setup up script for the APEL."""
+from shutil import copyfile
+from os import remove
+import sys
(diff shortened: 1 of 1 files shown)


tofu-rocketry:
I've added commits and rebased. Will use another PR when ready.

PR #96, opened 2016-04-04
Added files via upload

jcgordon:
Add star-accounting.py script to apel/scripts.so we have somewhere for users to pull modified version.

I will then make further changes.


tofu-rocketry:
This shouldn't be added to our main apel repository otherwise we'll need to maintain it.

PR #98, opened 2016-04-26
Remove args from create in logrotate script

### LINKED ISSUE #97: Logrotate assumes user 'apel' exists
Following from this GGUS ticket: https://ggus.eu/index.php?mode=ticket_info&ticket_id=120564

The logrotate script installed by the rpm assumes the presence of a user 'apel' on the system but the rpm installation does not create it. This leads to error messages in the file '/etc/cron.daily/logrotate':

error: apel-client:6 unknown user 'apel'
error: found error in /var/log/apel/client.log , skipping"

tofu-rocketry:
Resolves #97.

Remove the file attribute arguments from the create directive in the logrotate script as there isn't necessarily an apel user on the system. Without the file attributes defined, the create directive defaults to using the same values as for the original log file.

See description for `create mode owner group` in the [logrotate man page](http://www.linuxcommand.org/man_pages/logrotate8.html).

### DIFF
--- a/scripts/apel-client
+++ b/scripts/apel-client
@@ -3,5 +3,5 @@
     weekly
     rotate 4
     missingok
-    create 640 apel apel
+    create
 }


gregcorbett:
This looks fine

tofu-rocketry:
:tada:

PR #101, opened 2016-06-15
Torque core count fix

### LINKED ISSUE #100: Torque-5 change in core reporting
[GGUS ticket](https://ggus.eu/index.php?mode=ticket_info&ticket_id=121776).

> Torque-4.2.8 reports
> `exec_host=b372/4+b372/5+b372/6+b372/7`
> 
> while Torque-5.1.0 (and later) reports
>` exec_host=b391/0-3`
> and
> `total_execution_slots=4`
> (no such marker present in Torque-pre-5 logfiles)

tofu-rocketry:
Resolves #100.

- The `exec_host` field in Torque >= 5.1.0 gives things like
  `b391/0-1,5,11`,which is harder to parse, so for cores we can use the
  `total_execution_slots` field instead for the number of cores.
- Add tests to cover this.

### DIFF
--- a/apel/parsers/pbs.py
+++ b/apel/parsers/pbs.py
@@ -110,21 +110,28 @@ class PBSParser(Parser):
 
 
 def _parse_mpi(exec_host):
-    '''
+    """
     Return (nodes, total-cores) given a dict from a PBS record.
-
-    There are two ways to derive this from the logs; the second is from
-    the field
-    Resource_List.nodes=x:ppn=y
-    The above would return x,y.
-    '''
+    """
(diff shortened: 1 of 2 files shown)


tofu-rocketry:
Comments on test failures:

- Landscape is producing spurious results, again, as it's referring to files that haven't been changed in this PR.
- Codacy has picked up some overly long lines, but they're just long strings that I don't fancy having split over a huge number of lines.
- Codeclimate is being fussy about cyclomatic complexity.

gregcorbett:
Looks good :)

PR #105, opened 2016-08-26
Update Python versions in Travis file

tofu-rocketry:
- Remove Python 2.7 from allowed failures as we should now be building for
  OSes that have it as default.
- Update Python 3.4 to 3.5 as we can just check the latest available (could check all versions of 3.x, but that's overkill for a version that doesn't work for us at the moment).


gregcorbett:
looks good

PR #113, opened 2016-11-15
Add support for GOCDB paging

gregcorbett:
- Adds method to retrieve the `next` link from fetched XML (or `None` if at last page or paging not in use)
- Adds test case for this method
- Moves `get_xml()` into a while loop, which ends when the `next` link is `None`
- Adds a mock `gocdb_url` for the `test_basics()` test case, as this now needs to be set, otherwise the while loop ends immediately, without fetching XML
- Passes Document Object Model (DOM) between methods to prevent parsing the same XML twice. 
- Renames `dns_from_xml` to `dns_from_dom` accordingly.

### DIFF
--- a/test/test_retrieve_dns.py
+++ b/test/test_retrieve_dns.py
@@ -125,7 +125,7 @@ class RunprocessTestCase(unittest.TestCase):
 
         # Parse the test XML into a Document Object Model
         dom = xml.dom.minidom.parseString(xml_test_string)
-        
+
         result = bin.retrieve_dns.next_link_from_dom(dom)
         self.assertTrue(result, "next link")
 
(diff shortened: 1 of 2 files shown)


test/test_retrieve_dns.py
```diff
@@ -106,8 +106,29 @@ def setUp(self):
...
+        <SERVICE_ENDPOINT><HOSTDN>/basic/dn</HOSTDN></SERVICE_ENDPOINT>
+        <SERVICE_ENDPOINT><HOSTDN>/banned/dn</HOSTDN></SERVICE_ENDPOINT>
+        <SERVICE_ENDPOINT><HOSTDN>invalid</HOSTDN></SERVICE_ENDPOINT>
+        </results>"""
+
+        # Parse the test XML into a Document Object Model
+        dom = xml.dom.minidom.parseString(xml_test_string)
+
+        result = bin.retrieve_dns.next_link_from_dom(dom)
+        self.assertTrue(result, "next link")
```
tofu-rocketry: This should be `assertEqual`, should it not?

gregcorbett: It should have been! And now it is :)

PR #120, opened 2017-03-08
Cloud 0.4

### LINKED ISSUE #71: Implement V0.4 of the Cloud Schema
The updated cloud schema is described here:

https://oerc.basecamphq.com/F199381047

We need to implement:
- [x] Benchmark and BenchmarkType fields
- [x] subsite level
- [ ] Checking on ImageID field?

### LINKED ISSUE #112: Add cloud update schema in SPEC file

tofu-rocketry:
Finally resolves #71.
Also resolves #112.

Right, here it is at last! 🤞 

- This PR updates the cloud accounting to support the new schema v0.4 in terms of accepting that format of record and producing summaries from it. Also provided is an update schema for updating from a previous version of the schema.
- PublicIPCount is being excluded from the summaries as there is no agreement on how exactly this should be accounted for, whether as part of the VM record or as a separate resource.
- Currently missing is the inclusion of CPU counts in the cloud summaries - this will be added in a separate PR.

**N.B.** Double check that PublicIPCount only appears in cloud _records_, so that we can accept the record, but not _summaries_ (there a couple of commented out lines though).

(Note to self for future updates involving large changes in schema - keep whitespace changes to a minimum and tidy those up later as it makes the important changes more obvious! :grimacing: )

### DIFF
--- a/apel/db/records/cloud.py
+++ b/apel/db/records/cloud.py
@@ -38,24 +38,23 @@ class CloudRecord(Record):
         self._mandatory_fields = ["VMUUID", "SiteName"]
             
         # This list allows us to specify the order of lines when we construct records.
-        self._msg_fields  = ["VMUUID", "SiteName", "MachineName", 
+        self._msg_fields  = ["VMUUID", "SiteName", "CloudComputeService", "MachineName", 
                              "LocalUserId", "LocalGroupId", "GlobalUserName", "FQAN",
                              "Status", "StartTime", "EndTime", "SuspendDuration", 
                               "WallDuration",
                              "CpuDuration", "CpuCount", "NetworkType", "NetworkInbound", 
-                             "NetworkOutbound",
-                             "Memory", "Disk", "StorageRecordId", "ImageId", "CloudType"]
+                             "NetworkOutbound", "PublicIPCount", "Memory", "Disk", 
+                             "BenchmarkType", "Benchmark", "StorageRecordId", "ImageId", "CloudType"]
         
         # This list specifies the information that goes in the database.
-        self._db_fields = self._msg_fields[:7] + ['VO', 'VOGroup', 'VORole'] + self._msg_fields[7:]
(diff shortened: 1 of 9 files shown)


gregcorbett:
Looks good to me. And it looks like you've removed all the references to PublicIPCount in the summaries (and in the test cases!) that were introduced.

tofu-rocketry:
Ah, I forgot to double check the number of arguments to the procedure calls in `mysql.py`. With the removal of `PublicIPCount` there's now one too many for `ReplaceCloudSummaryRecord`.

gregcorbett:
good spot!

apel/db/records/cloud.py
```diff
@@ -38,25 +38,25 @@ def __init__(self):
...
-                             "NetworkOutbound",
-                             "Memory", "Disk", "StorageRecordId", "ImageId", "CloudType"]
+                             "WallDuration", "CpuDuration", "CpuCount", 
+                             "NetworkType", "NetworkInbound", "NetworkOutbound", "PublicIPCount", 
+                             "Memory", "Disk", "BenchmarkType", "Benchmark", 
+                             "StorageRecordId", "ImageId", "CloudType"]
         
         # This list specifies the information that goes in the database.
-        self._db_fields = self._msg_fields[:7] + ['VO', 'VOGroup', 'VORole'] + self._msg_fields[7:]
+        self._db_fields = self._msg_fields[:8] + ['VO', 'VOGroup', 'VORole'] + self._msg_fields[8:]
```
tofu-rocketry: Would you agree that this section still holds true with the removal of `PublicIPCount`? The split is before the point where it was added.

gregcorbett: yes, because FQAN has moved from index 7 to 8 with the addition of `CloudComputeService` so changing `7` to `8` at line 50 means the VO fields stay in the same place

tofu-rocketry: Thanks. I "dismissed" your review when I realised there was a mistake so could you re-review, please?

PR #123, opened 2017-03-16
Change Update Timestamp to be explicit

### LINKED ISSUE #117: UpdateTime in CloudRecords is never set
The UpdateTime of the CloudRecords table is not set when messages are loaded.

This is because it is defined in`schemas/cloud.sql` as `UpdateTime TIMESTAMP`.

When deployed from scratch it defaults to `UpdateTime TIMESTAMP NULL DEFAULT NULL`. 

As such it not set when the Record gets created/updated

It should be set to something like: `UpdateTime TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP`

### LINKED ISSUE #118: UpdateTime in JobRecords/Summaries is never set
The UpdateTime of the JobRecords table is not set when messages are loaded.

This is because it is defined in `schemas/server.sql` as `UpdateTime TIMESTAMP`.

When deployed from scratch it defaults to `UpdateTime TIMESTAMP NULL DEFAULT NULL`.

As such it not set when the Record gets created/updated

It should be set to something like: `UpdateTime TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP`

The Summaries table also has the same problem.

gregcorbett:
Resolves #117, and resolves #118.

In MySQL 5.1, 

> The first `TIMESTAMP` column in a table, if not declared with the `NULL` attribute or an explicit `DEFAULT` or `ON UPDATE` clause, is automatically assigned the `DEFAULT CURRENT_TIMESTAMP` and `ON UPDATE CURRENT_TIMESTAMP` attributes.

In MySQL 5.6.6+, this does not happen. 

> No `TIMESTAMP` column is assigned the `DEFAULT CURRENT_TIMESTAMP` or `ON UPDATE CURRENT_TIMESTAMP` attributes automatically. Those attributes must be explicitly specified."

See [here](https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_explicit_defaults_for_timestamp) for the MySQL documentation referring to this. 

This PR explicitly specifies `NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP` for all columns called `UpdateTime`, and provides an `update_schema.sql` file to upgrade any existing databases. Any `NULL` time stamps will be set to `CURRENT_TIMESTAMP` when `update_schema.sql` is applied.

Once this PR is applied, `UpdateTime` will now explicitly update on 5.6.6+ as it does on implicitly on 5.1


tofu-rocketry:
I've edited it now, but note that PR messages need closing keywords before each PR number.

gregcorbett:
ah okay. I see what you mean

scripts/update_timestamp_explicit_client.sql
```diff
@@ -0,0 +1,13 @@
+-- UPDATE SCRIOT FOR CLIENT SCHEMA
```
tofu-rocketry: "SCRIPT"

scripts/update_timestamp_explicit_client.sql
```diff
@@ -0,0 +1,13 @@
...
+-- This script will set any null UpdateTimes to now
```
tofu-rocketry: As we use `UpdateTime` for decisions about how recent a record is, setting NULLs to now has the potential to cause problems.

I think it would be better to set any NULLs (which there shouldn't be any of hopefully) to '0000-00-00 00:00:00' (the “zero” timestamp) in all the update scripts. Thoughts?

scripts/update_timestamp_explicit_client.sql
gregcorbett: Done

scripts/update_timestamp_explicit_client.sql
gregcorbett: Sounds good, I'll make the change

gregcorbett: For just the client schema? or for all schema (i.e. `server.sql`, `cloud.sql`, etc)?

tofu-rocketry: For all.

gregcorbett:
Changes made

tofu-rocketry:
Thanks. Can you rebase? Then I'll merge.

gregcorbett:
rebased

PR #125, opened 2017-03-20
Add extra PBS test case to cover odd processors

tofu-rocketry:
- Add extra PBS test to cover the case where more than one processor is
  requested but only one seems to be assigned and there is no
  total_execution_slots in the output. This is hopefully no longer a
  common case with the addition of the execution slots in later versions.
  It was never clear exactly how this situation should be handled.


tofu-rocketry:
This test should perhaps fail, or at least the parser should possibly raise a warning for the user.

PR #126, opened 2017-03-20
Replace backslashes with implicit line continuation

tofu-rocketry:
Replace backslashes used for line continuation in Python code with implicit line continuation using parentheses as the former style is discouraged and is fragile.

There are places that use backslashes in shell code, but that's a separate issue.

### DIFF
--- a/apel/common/datetime_utils.py
+++ b/apel/common/datetime_utils.py
@@ -104,12 +104,12 @@ def iso2seconds(isoduration):
             intvals.append(0)
 
     # ignore group 0 ("P") and group 5 ("T")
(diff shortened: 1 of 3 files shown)


gregcorbett:
There are some back slashes in test/test_record_factory.py (line number 68 and 106). 

Do you want to get rid of them as well?

tofu-rocketry:
Good spot (I didn't), but I'm not too concerned about backslashes in strings as they perform a slightly different function. In this case they are, in effect, removing an implicit newline character so that the string can be formatted as it would appear in an APEL job message.

PR #131, opened 2017-04-19
Changes to make APEL parser process and send accounting for HTCondorCE

jcasals:
New pull request fixing the error in the conf file. These are the modified files:

```
apel/db/backends/mysql.py
apel/db/records/__init__.py
apel/parsers/htcondor.py
conf/parser.cfg
apel/db/records/htcondorce.py
apel/parsers/htcondorce.py
apelparser
```

### DIFF
--- a/apel/db/records/htcondorce.py
+++ b/apel/db/records/htcondorce.py
@@ -20,23 +20,22 @@ class HTCondorCERecord(Record):
     """
     Class to represent a record from BLAHP systems
     """
-    
+
     def __init__(self):
         '''
         Initializer for BlahdRecord
         '''
-        
+
         Record.__init__(self)
-        
+
         self._mandatory_fields = []
-        
+
         self._db_fields = ["TimeStamp", "GlobalUserName", "FQAN",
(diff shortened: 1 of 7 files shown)


geonmo:
@jcasals Could you fix this PR asap? I need to build our EGI site in May.

gregcorbett:
@jcasals apologies for the delay in getting back to you on this.

Could you explain why `apel/parsers/htcondor.py` had to be changed to get the CE functionality working?

We had assumed the new CE parser would have been completely separate from the existing HTCondor parser we have.

kreczko:
@gregcorbett Since HTCondor CE is just a special configuration of an HTCondor Scheduler wouldn't it make sense to use the same code?

apel/db/records/htcondorce.py
```diff
@@ -0,0 +1,41 @@
+'''
+   Copyright 2012 Konrad Jopek
```
kjopek: I guess I am not author of this file :)

tofu-rocketry: I'll sort out all the notices before anything's merged to dev.

PR #133, opened 2017-04-25
Add a setup script

### LINKED ISSUE #86: Add setup.py script

gregcorbett:
Resolves #86 

Adds a setup.py script to APEL, allows APEL packages to be installed on any OS that has Python installed

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -79,6 +79,9 @@ def main():
           download_url='https://github.com/apel/apel/releases',
           license='Apache License, Version 2.0',
           install_requires=['MySQL-python', 'iso8601', 'python-ldap', 'dirq'],
+          extras_require={
+              'python-daemon': ['python-daemon'],
+          },
           packages=find_packages(exclude=['bin']),
           scripts=['bin/apelclient',
                    'bin/apelparser',
--- a/setup.py
+++ b/setup.py
@@ -66,7 +66,7 @@ def main():
     log_rotate_dir = '/etc/logrotate.d'
     log_rotate_files = ['scripts/apel-client']
 
-    # For 'python setup.py install' to 
+    # For 'python setup.py install' to
     # work (on Linux SL6), 'python-daemon'
     # must be installed or included
     # in install_required
--- a/setup.py
+++ b/setup.py
@@ -97,7 +97,9 @@ def main():
                       (log_rotate_dir, log_rotate_files)],
           # zip_safe allows setuptools to install the project
           # as a zipfile, for maximum performance!
-          zip_safe=True)
+          # We have disabled this feature so installing via the setup
+          # script is similar to installing the RPM apel-ssm
+          zip_safe=False)
 
     # Remove temporary files with deployment names
     if 'install' in sys.argv:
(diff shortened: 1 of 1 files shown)


tofu-rocketry:
If an issue exists for a pull request, I prefer not to add a label to the pull request.

setup.py
```diff
@@ -0,0 +1,107 @@
+"""A setup script for APEL.
+
+This script installs the APEL library, client, parsers and server. This
+should be similar to installing the RPMs for apel-lib, apel-client,
+apel-parsers, and apel-server, although there may be some differences.
```
tofu-rocketry: Does this not install the .pyc/o like the SSM setup.py? If so, should use a similar comment to SSM's.

setup.py
```diff
@@ -0,0 +1,107 @@
...
+                   'bin/apelauth'],
+          data_files=[(conf_dir, conf_files),
+                      (data_dir, schema_files),
+                      (data_dir, accounting_files),
+                      (data_dir, message_files),
+                      (data_dir, update_scripts),
+                      (log_rotate_dir, log_rotate_files)],
+          # zip_safe allows setuptools to install the project
+          # as a zipfile, for maximum performance!
+          zip_safe=True)
```
tofu-rocketry: Let's stick this to `False` like ssm's setup.py.

setup.py
```diff
@@ -0,0 +1,107 @@
...
+
+    accounting_files = ['scripts/slurm_acc.sh', 'scripts/htcondor_acc.sh']
+
+    message_files = ['scripts/msg_status.py']
+
+    # log_rotate_files will later be copied to log_rotate_dir
+    log_rotate_dir = '/etc/logrotate.d'
+    log_rotate_files = ['scripts/apel-client']
+
+    # For 'python setup.py install' to
```
tofu-rocketry: Should we have an `extras_require` like for ssm?

tofu-rocketry: This did appear in the version before the rebase - did it get lost?

gregcorbett: probably, can you post a link to the pre-rebase version?

tofu-rocketry: Not to that exactly as I only have it locally, but it should be the same as for SSM:

https://github.com/apel/ssm/blob/dev/setup.py#L61

setup.py
gregcorbett: it does not, comment added

setup.py
gregcorbett: added

setup.py
gregcorbett: done

gregcorbett:
changes made

gregcorbett:
the inability to use `python setup.py test` does seem to be due to the use of `..` in `test/test_schemas.py`. presumably the tests aren't being run from the `test` directory, but rather the repo root, where `setup.py` is.

running `python ../setup.py test` from the `test` directory does successfully run the test suite, if one re-enables the test functionality of `setup.py`.

tofu-rocketry:
Looks good. Could you rebase? Then I'll review and merge.

Issue #128, opened 2017-03-30
Record._check_fields() does not check float or datetime fields


gregcorbett:
This leads to `'None'` being passed to the database, rather than a correctly typed value (or `NULL`)

This is allowed (by default) in MySQL 5.1 and rather than fail it interprets `'None'` as the zero value for that data type (i.e. `0.00` for floats and `00/00/00 00:00:00` for datetimes). 

In MySQL 5.6.X this is not the default, as such records missing a datetime or float field will not be loaded.

gregcorbett:
I've created a [new branch](https://github.com/gregcorbett/apel/tree/type_checking) with a [new test](https://github.com/gregcorbett/apel/blob/type_checking/test/test_cloud_record.py#L189) to illustrate [the issue](https://travis-ci.org/gregcorbett/apel/jobs/216764198). 

It uses the cloud test record because that was where the problem was first noticed, but it is probaly present in all records, as they all inherit `_check_fields()` from `Record`.

tofu-rocketry:
What was the MySQL setting that you said was causing this?

gregcorbett:
`sql_mode`, in 5.1 it is set to `''` and in 5.6.X it is set to `STRICT_TRANS_TABLES,NO_ENGINE_SUBSTITUTION`.

Globally (and persistently with respect to mysql server restarts) setting `sql_mode` to `''` in 5.6.X is a short term fix.

tofu-rocketry:
MariaDB 10.1.19 doesn't have strict set, though later versions do.

gregcorbett:
Shall I add similar tests to `test_load_from_msg_type_check` for Storage and Job Records?

gregcorbett:
Having tried to implement them I don't think they are worth it. 

The test checks that missing values get correctly typed, such as `Benchmark` when we get sent a Cloud v0.2 record (so as not to rely on MySQL type casting). There isn't a similar case for Storage and Grid, but new tests could be based off of `CloudRecordTest.test_load_from_msg_type_check` when the Record format update happens.

PR #135, opened 2017-05-25
Add a test for loading cloud messages (from the APEL message format) to the database

gregcorbett:
This test, whilst passing and not adding to the code coverage, does reveal database warnings that become failures with stricter `sql_mode` settings, as referenced in #128.

This test will ensure the fix for #128 does not break the cloud loader, i.e by trying to pass a `NULL` benchmark to the database while `Benchmark` is a `NOT NULL` field), which is currently not caught by the unit tests.

Changes to `apel/db/backends/mysql.py`:
- will hopefully allow future equality testing of coud records pre and post database load in a similar fashion to job records by allowing `_get_records` to work for cloud records  
- it seems like `_get_records` only used in tests (i.e. `MYSQL_TABLES` is only mentioned in
  `_get_records`/`get_records`, which are in turn only called in testing

### DIFF
--- a/apel/db/backends/mysql.py
+++ b/apel/db/backends/mysql.py
@@ -49,7 +49,7 @@ class ApelMysqlDb(object):
                     JobRecord   : 'VJobRecords',
                     BlahdRecord : 'BlahdRecords',
                     SyncRecord  : 'SyncRecords',
-                    CloudRecord : 'CloudRecords',
+                    CloudRecord : 'VCloudRecords',
                     CloudSummaryRecord : 'VCloudSummaries',
                     NormalisedSummaryRecord : 'VNormalisedSummaries',
                     ProcessedRecord : 'VProcessedFiles',
--- a/schemas/cloud.sql
+++ b/schemas/cloud.sql
@@ -211,7 +211,7 @@ DELIMITER ;
 
 -- -----------------------------------------------------------------------------
 -- Sites
-
+DROP TABLE IF EXISTS Sites;
 CREATE TABLE Sites (
     id INT NOT NULL AUTO_INCREMENT PRIMARY KEY
  ,  name VARCHAR(255) NOT NULL
@@ -237,7 +237,7 @@ DELIMITER ;
 
 -- -----------------------------------------------------------------------------
 -- CloudComputeService
-
+DROP TABLE IF EXISTS CloudComputeServices;
 CREATE TABLE CloudComputeServices (
     id INT NOT NULL AUTO_INCREMENT PRIMARY KEY
  ,  name VARCHAR(255) NOT NULL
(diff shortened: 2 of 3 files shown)


gregcorbett:
I've just tested a `loader` run with 16,000 previously accepted messages an no rejects caused by this change :)

apel/db/backends/mysql.py
```diff
@@ -49,7 +49,7 @@ class ApelMysqlDb(object):
                     JobRecord   : 'VJobRecords',
                     BlahdRecord : 'BlahdRecords',
                     SyncRecord  : 'SyncRecords',
-                    CloudRecord : 'CloudRecords',
```
tofu-rocketry: Little surprising this has hung about for so long!

schemas/cloud.sql
```diff
@@ -211,7 +211,7 @@ DELIMITER ;
 
 -- -----------------------------------------------------------------------------
 -- Sites
-
+DROP TABLE IF EXISTS Sites;
```
tofu-rocketry: Good to have this inconsistency sorted out. :) Pretty much all the other tables have a `DROP TABLE`.

test/test_mysql.py
```diff
@@ -149,6 +204,57 @@ def test_last_update(self):
...
+CloudType: OpenNebula
+'''
+
+CLOUD4 = '''VMUUID: 2012-08-14 14:00:01+0200 FZJ Accounting Test
+SiteName: FZJ
+CloudComputeService: OpenStack Service A
+MachineName: Accounting Test
+LocalUserId: 1189105086dc4959bc9889383afc43b5
+LocalGroupId: EGI FCTF
+GlobalUserName: /DC=es/DC=irisgrid/O=cesga/CN=javier-lopez
```
tofu-rocketry: Records in tests should be anonymised.

test/test_mysql.py
```diff
@@ -79,6 +79,61 @@ def test_load_and_get(self):
         # Can't use 'all()' rather than comparing the length as Python 2.4
         self.assertEqual([item in items_out for item in items_in].count(True), len(items_in))
 
+    def test_load_and_get_cloud(self):
+        '''
+        Test a CloudV0.2/0.4 message can be loaded into the database.
+
+        It currently can't test for pre/post database load equality
+        as the two are currently not the same for V0.2.
+        i.e Benchmark 'None', which gets saved a set to 0.0
```
tofu-rocketry: "i.e." and either "gets saves as" or "gets set to".

gregcorbett:
changes made

tofu-rocketry:
Thanks. Can you do an interactive rebase and squash 1fdf300 into 5a02c2a so that the non-anonymised version isn't kept?

gregcorbett:
Done

PR #136, opened 2017-05-31
Add Long Running VM support (version 2)

gregcorbett:
Assigns usage to a month based on a `RecordCreateTime` or a measurement time.

If a VM has completed, it's Measurement Time is its `EndTime`, if a VM has not completed then either its `RecordCreateTime` is used as it's Measurement Time or the Measurement Time is calculated as `StartTime + IFNULL(WallDuartion,0) + IFNULL(SuspendDuration,0)`. If that calculation fails (because of a lack of `StartTime`) the usage gets assigned to 0 month/0 year, which is what happens currently with such records.

If the `RecordCreateTime` is not present in the message it will be set to the Measurement Time

Usage now gets assigned to the `MeasurementMonth`/`MeasurementYear` rather than `Month(StartTime)`/`Year(StartTime)`, otherwise summarise remain unchanged.


schemas/cloud.sql
```diff
@@ -72,18 +79,103 @@ CREATE PROCEDURE ReplaceCloudRecord(
...
+            SET measurementTimeCalculated = IFNULL(measurementTimeCalculated, '00-00-00 00:00:00');
+            SET recordCreateTimeNotNull = measurementTimeCalculated;
+        ELSE
+             -- Use the supplied record create time as the measurement time
+            SET measurementTimeCalculated = recordCreateTime;
+            SET recordCreateTimeNotNull = recordCreateTime;
+        END IF;
+    END IF;
+
+    INSERT INTO CloudRecords(RecordCreateTime, VMUUID, SiteID, CloudComputeServiceID, MachineName,
```
tofu-rocketry: Why the change from REPLACE INTO?

gregcorbett: you can't use `ON DUPLICATE KEY UPDATE` with a `REPLACE`.

So using `REPLACE` would mean the record incoming record always takes precedence over the existing one. But by using `ON DUPLICATE KEY UPDATE` you can tell it to use the one with the highest `MeasurementTime`

gregcorbett:
thanks, will do

schemas/cloud.sql
tofu-rocketry: Ah, ok.

gregcorbett:
PR is up to date with development/testing work

gregcorbett:
I have rebased this atop 1.6.1 and made the changes to the update script.

Test loading of previously accepted message from Feb17 onwards was successful (no rejects). As before, most sites continue to have agreement between the prod database and the database post update and reprocessing. The sites that don't are the same sites as before and those site were ticketed then for help.

gregcorbett:
#156 isn't a blocking issue for this, I have tested the LRVM code without applying the bug fix and it had no effect on the warnings.

I would say get LRVMs out the way and we can fix the CpuCount warnings and Month Year warnings at a later date.

gregcorbett:
rebased

PR #138, opened 2017-06-28
Rename Update Scripts so it is clearer what versions they apply to

gregcorbett:
A proposal on how to better manage the update schema scripts we have been accumulating.

`update_schema.sql` has been renamed to `update_1.5.1_to_1.6.0.sql`. It's more descriptive and it's immediately clear what version it should be applied to.

The update scripts we have introduced since the last tagged version have been combined into `update_1.6.0_to_next_tagged_version.sql`, again I think this makes it immediately clear what version of APEL the updates should be applied to.


gregcorbett:
changes made. landscape errors seem to be highlighting problems in `migrate_apel.py`, which I have not touched....

tofu-rocketry:
Yeah, I'm thinking of disabling Landscape as it's quite hit and miss whether it even runs!

I've push a couple of commits. Looks good otherwise. If you're happy with my changes, please rebase.

gregcorbett:
rebased

PR #139, opened 2017-08-23
Prevent compiled python files from being installed by the RPM

gregcorbett:
- Replace wildcard match with explicit excluding of `.pyc`/`.pyo` files

- The `.pyc`/`.pyo` files are auto-generated by `rpmbuild` so we can't stop them being generated and we cant just remove the wildcard because `rpmbuild` complains the `.pyc`/`.pyo` are present but not added to the RPM. So we need to explicitly exclude them from the RPM.

- 'exclude' is not the most documented spec file directive

### DIFF
--- a/apel.spec
+++ b/apel.spec
@@ -177,8 +177,9 @@ exit 0
 %_datadir/apel/update-1.5.1-1.6.0.sql
 %_datadir/apel/update-1.6.0-next.sql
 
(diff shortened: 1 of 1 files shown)


tofu-rocketry:
Looks good. I presume you've tested this?

If so, can you rebase, please?

gregcorbett:
yes, it has been tested on SL6 and SL7. 

Rebased.

PR #140, opened 2017-08-23
Remove LSF version restriction

### LINKED ISSUE #129: Add support for LSF v10
From [GGUS](https://ggus.eu/index.php?mode=ticket_info&ticket_id=126319).

Changes may be minimal considering that LSF has been well behaved in its accounting since version 5.

Should also consider adding forwards compatibility.

tofu-rocketry:
Resolves #129.

- Remove LSF version restriction from LSF parser (where a TypeError would be raised if there was an unexpected version in the accounting log) as LSF has been well behaved over a number of major versions, it causes problems for users trying to upgrade to newer versions that haven't been tested, and we don't check versions in any other parser. This should remove the burden of updating the version mapping whenever a new LSF version is released.
- Remove unittest for invalid LSF version as version is no longer checked.

Site from GGUS ticket (see issue #129) has used this updated parser successfully.

### DIFF
--- a/apel/parsers/lsf.py
+++ b/apel/parsers/lsf.py
@@ -108,34 +108,27 @@ class LSFParser(Parser):
         else:
             nnodes = 0
             ncores = 0
-        
-        mapping_lsf = {'Site'          : lambda x: self.site_name,
-                       'JobName'       : lambda x: x[3],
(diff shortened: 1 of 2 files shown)


tofu-rocketry:
Coverage reduced a tiny bit as two covered lines were completely removed.

PR #141, opened 2017-08-23
Add type checking to float and datetime fields of records

### LINKED ISSUE #128: Record._check_fields() does not check float or datetime fields

gregcorbett:
Resolves #128 

`Record._check_fields()` now checks fields listed in `self._float_fields` and `self._datetime_fields`  are the appropriate type, i.e. `float` or `datetime` in a similar fashion to how `self._int_fields` are currently handled. In the case of `self._float_fields` it will try and cast the input to a `float` if it is not already.

This results in properly typed fields or `None` to be passed to the database, rather than properly typed fields or the string `'None'`. 

This fixes the problem that MySQL 5.6.X has strict type checking and wont do the type casts that 5.1.X would.

The change to `apel/db/records/job.py` is due to the fact the `_check_factor` method is being passed `None` if ServiceLevel is missing (rather than the string `'None'`). The output of `_check_factor` is not changed.

The test included in this PR stop short of actually trying to load the messages. As such this should be merged in after #135 , as that PR provides a test that will confirm this doesn't break the loader.

### DIFF
--- a/apel/db/records/job.py
+++ b/apel/db/records/job.py
@@ -146,12 +146,12 @@ class JobRecord(Record):
         Check for the validity of the ScalingFactorUnit and ScalingFactor fields.
         We accept neither field included or both.  If only one of the fields is 
         included, it doesn't really make sense so we reject it.
-    
+
         We expect that:
-           - As ServiceLevel is a float field,
-             a null sf has been converted to None.
-           - As ServiceLevelType is a msg field,
-             a null sf has been converted to the string 'None'.
+        - as ServiceLevel is a float field,
+          a null ScalingFactor is passed in as the null object None.
+        - as ServiceLevelType is a msg field,
+          a null ScalingFactorUnit is passed in as the string 'None'.
         '''
         if sf == None:
             if sfu != 'None':
--- a/apel/db/records/job.py
+++ b/apel/db/records/job.py
@@ -147,9 +147,13 @@ class JobRecord(Record):
         We accept neither field included or both.  If only one of the fields is 
         included, it doesn't really make sense so we reject it.
     
-        We expect that all null values have been converted to the string 'None'.
+        We expect that:
+           - As ServiceLevel is a float field,
+             a null sf has been converted to None.
+           - As ServiceLevelType is a msg field,
+             a null sf has been converted to the string 'None'.
         '''
-        if sf == 'None':
+        if sf == None:
             if sfu != 'None':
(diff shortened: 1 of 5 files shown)


tofu-rocketry:
#128 is focussed on cloud accounting, but what is the impact of these changes on grid and storage accounting?

gregcorbett:
I don't think there is any impact as I don't believe we are sent old versions of the Grid record (and older versions of Storage Records dont exist?), so we arent sent records that are missing `ServiceLevel` for example

gregcorbett:
rebased

apel/db/records/job.py
```diff
@@ -146,10 +146,14 @@ def _check_factor(self, sfu, sf):
         Check for the validity of the ScalingFactorUnit and ScalingFactor fields.
         We accept neither field included or both.  If only one of the fields is 
         included, it doesn't really make sense so we reject it.
-    
-        We expect that all null values have been converted to the string 'None'.
+
+        We expect that:
```
tofu-rocketry: I changed these to make it clearer that we're expecting the right sort of arguments to the method.

test/test_cloud_record.py
```diff
@@ -164,16 +165,16 @@ def setUp(self):
...
                          'CloudType': 'OpenNebula'}
         self.cases = {}
         self.cases[self._msg1] = self._values1
         self.cases[self._msg2] = self._values2
         self.cases[self._msg3] = self._values3
 
-    def test_load_from_msg(self):
-        
+    def test_load_from_msg_value_check(self):
+        """Check for correct values in CloudRecords generated from messages."""
```
tofu-rocketry: I've changed this to be more about what we want to happen, rather than a fairly direct description of the code. Tests and their comments/docstrings can be quite useful when you need to check what _should_ be happening.

tofu-rocketry:
Could you have a look through my changes, and, if you're happy, rebase, please?

gregcorbett:
rebased

gregcorbett:
and happy :)

PR #142, opened 2017-08-24
Fixes to XML parsers

tofu-rocketry:
- Main fix is to `xml_parser.py`. The `getTagByAttr` method in that should have been using the namespace it's given as an argument throughout (so that you can choose the namespace when you call it) but it was using `self.NAMESPACE` for one of the calls it makes and thus overriding your choice. This is the most significant change as it affects all the other XML parsers (StAR, CAR, AUR), but doesn't seem to have been a problem for the StAR parser as it uses a uniform namespace.
- Added a regression test for this.
- Tidied up the CAR parser.
- Fixed the AUR parser somewhat. It still doesn't work out normalised durations, but can now get Infrastructure, VOGroup, and VORole. (It's never really worked, to be fair! It's even disabled in `record_factory.py`.)
- Fixed the test for the AUR parser.

### DIFF
--- a/apel/db/loader/aur_parser.py
+++ b/apel/db/loader/aur_parser.py
@@ -45,7 +45,7 @@ class AurParser(XMLParser):
     
     def get_records(self):
         '''
-        Returns list of parsed records from CAR file.
+        Returns list of parsed records from AUR file.
         
         Please notice that this parser _requires_ valid
         structure of XML document, including namespace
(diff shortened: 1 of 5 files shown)


gregcorbett:
the AUR record is the XML based Grid Summary Record isn't it?

tofu-rocketry:
Yes, though it's actually not enabled in `record_factory.py`. This was more an exercise to fix up the bug that was in `getTagByAttr` and brush up my Python-based XML parsing. Would be nice sometime to get the parser properly conformant to the EMI XML record format, and eventually to the OGF format.

tofu-rocketry:
Would it be ok to merge this before #141 is sorted out?

gregcorbett:
yes

PR #144, opened 2017-09-07
Minor fixes and improved coverage

tofu-rocketry:
This fixes a couple of very minor things that aren't even really bugs, and is mainly adding extra unittests to the ones that were straightforward to do. I'll just list the non-test changes here:
- Fix spacing and capitalisation of constructed queries in `apeldb` module to be more consistent and match standard SQL.
- Add `HTCondorParser` to `__init__` imports.

### DIFF
--- a/test/test_car_parser.py
+++ b/test/test_car_parser.py
@@ -187,10 +187,16 @@ class CarParserTest(unittest.TestCase):
 
     def test_empty_xml(self):
         """
-        Check that correct exception is raised for an XML file with no records.
+        Check that exception is raised for XML not in computerecord namepsace.
         """
         parser = CarParser("<something></something>")
         self.assertRaises(XMLParserException, parser.get_records)
+
+    #def test_empty_record(self):
+    #    # Not sure what should be returned with an empty record like this.
+    #    parser = CarParser('<urf:UsageRecord xmlns:urf="http://eu-emi.eu/namesp'
+    #                       'aces/2012/11/computerecord"></urf:UsageRecord>')
+    #    print parser.get_records()[0]._record_content
 
 if __name__ == '__main__':
     unittest.main()
--- a/test/test_car_parser.py
+++ b/test/test_car_parser.py
@@ -2,6 +2,7 @@ import datetime
 import unittest
 
 from apel.db.loader import CarParser
+from apel.db.loader.xml_parser import XMLParserException
 
 
 def datetimes_equal(dt1, dt2):
@@ -183,6 +184,13 @@ class CarParserTest(unittest.TestCase):
                         self.fail("Datetimes don't match for key %s: %s, %s" % (key, cont[key], cases[car][key]))
                 else:
                     self.assertEqual(cont[key], cases[car][key], '%s != %s for key %s' % (cont[key], cases[car][key], key))
+
+    def test_empty_xml(self):
+        """
+        Check that correct exception is raised for an XML file with no records.
+        """
+        parser = CarParser("<something></something>")
+        self.assertRaises(XMLParserException, parser.get_records)
 
 if __name__ == '__main__':
     unittest.main()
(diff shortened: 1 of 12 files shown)


test/test_parsers.py
```diff
@@ -0,0 +1,89 @@
...
+                '1007|c_61|dt5|dtm|2013-03-27T17:13:41|2013-03-27T17:13:44|00:0'
+                '0:03|3|prod|1|1|c-40|||COMPLETED'),
+        }
+
+        for test in tests:
+            parser_object = getattr(apel.parsers, test + 'Parser')
+            parser = parser_object('testSite', 'testHost', False)
+            self.assertTrue(parser.recognize(tests[test]), test)
+
+    def test_none(self):
```
gregcorbett: Why doesn't this test method test the Blah, HTCondor or SGE parser? Like `test_good` does?

test/test_car_parser.py
```diff
@@ -184,5 +185,12 @@ def test_car_parser(self):
                 else:
                     self.assertEqual(cont[key], cases[car][key], '%s != %s for key %s' % (cont[key], cases[car][key], key))
 
+    def test_empty_xml(self):
+        """
+        Check that correct exception is raised for an XML file with no records.
+        """
+        parser = CarParser("<something></something>")
```
gregcorbett: I feel like this doesn't just test for a XML file with no records, but just checks that garbage XML raises the exception.

Would the message below also raise the exception?
```
<?xml version="1.0" encoding="UTF-8"?>
<urf:UsageRecord xmlns:urf="http://eu-emi.eu/namespaces/2012/11/computerecord" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://eu-emi.eu/namespaces/2012/11/computerecord ../../../org.glite.dgas_B_4_0_0/config-consumers/car_v1.2.xsd ">
</urf:UsageRecord>
```

test/test_record_factory.py
```diff
@@ -23,27 +24,67 @@ def test_create_records(self):
...
+        """Check that creating a StAR record returns a StorageRecord."""
+        record_list = self._rf.create_records(
+            '<sr:StorageUsageRecord xmlns:sr="http://eu-emi.eu/namespaces/2011/'
+            '02/storagerecord"><sr:RecordIdentity sr:createTime="2016-06-09T02:'
+            '42:15Z" sr:recordId="c698"/></sr:StorageUsageRecord>'
+        )
+        self.assertTrue(isinstance(record_list[0], StorageRecord))
+
+    def test_create_bad_xml(self):
+        """Check that trying to create a record from non-record XML fials."""
```
gregcorbett: should that be 'fails'?

tofu-rocketry:
Yeah, I'll docstring them. I got lazy.

test/test_record_factory.py
tofu-rocketry: Yes, it should!

test/test_parsers.py
tofu-rocketry: Ah, the docstring kind of explains it, but it is a little unclear.

Only these parsers have the additional functionality to return `None` when they parse a syntactically valid line that should be ignored for some reason (e.g. unfinished). All the others either return the record or raise an exception.

I'll update the docstring if that's a better explanation.

test/test_car_parser.py
tofu-rocketry: I'll clarify the wording in that test.

That message doesn't raise an exception. Not sure what it should do, but I don't want to dive down the car_parser rabbit hole right now. I'll add a commented out test for it though.

test/test_parsers.py
gregcorbett: okay

test/test_car_parser.py
gregcorbett: looks good

gregcorbett:
looks good

tofu-rocketry:
My first all green PR is ages! (Silly code quality checks. 😛 )

PR #145, opened 2017-09-11
Storage fixes

### LINKED ISSUE #108: Update storage views
Following on from #106 the views in the storage schema needs to be updated to include the extra two fields.

~~Additionally, the anonymous views being used to send the data to the portal are not included in the code in git and should perhaps be added here.~~

Previous sentence is incorrect - it is actually the VStarRecords view that is being used, but this was obscured by the separate setup that the old pre production machine had. So the anonymous views don't seem to serve a purpose (testing, perhaps?) and could be skipped.

Additionally, the unloader code doesn't import the StarRecord and doesn't list the table name, so can't actually unload using the RPM installed version. This was handled by a manually altered unloader that was used by changing the Python path before starting the unloader.

### LINKED ISSUE #109: Add storage update schema
Similar to the updates to the cloud schema, storage should have an update schema to update it to the latest version.

tofu-rocketry:
Resolves #108, resolves #109.

Changes in this PR:

- Add bits to unloader to enable unloading of StAR records in XML format.
- Add `VStarRecords` view.
- Extend the two update schemas to enable migration of old storage accounting databases.

The update schema changes were tested by creating a database for each version (1.5.1, 1.6.0, 1.6.1), running the relevant update script on them, then dumping the database and diffing the output (ignoring whitespace).

```
# mysqldump --skip-comments --skip-extended-insert --routines -u root v151 > v151.sql

# diff -w v151.sql 161new.sql
```

No significant differences between all three, though the two databases that had been updated had the additional default values in the `Roles` and `SubGroups` tables (as per the update schema) and so, interestingly, the table properties for them included `AUTO_INCREMENT=2` which sets the initial `AUTO_INCREMENT` value for the table to 2 (one more than the default value).

### DIFF
--- a/apel/db/records/storage.py
+++ b/apel/db/records/storage.py
@@ -175,6 +175,18 @@ class StorageRecord(Record):
             group_node.appendChild(doc.createTextNode(group_field))
             s_identity.appendChild(group_node)
 
+        if self.get_field('SubGroup') is not None:
+            sub_attr = doc.createElement('sr:GroupAttribute')
+            sub_attr.setAttribute('sr:attributeType', 'subgroup')
+            sub_attr.appendChild(doc.createTextNode(self.get_field('SubGroup')))
+            s_identity.appendChild(sub_attr)
+
+        if self.get_field('Role') is not None:
+            role_attr = doc.createElement('sr:GroupAttribute')
+            role_attr.setAttribute('sr:attributeType', 'role')
+            role_attr.appendChild(doc.createTextNode(self.get_field('Role')))
+            s_identity.appendChild(role_attr)
+
         # Append Subject Identity Block
         ur.appendChild(s_identity)
 
--- a/apel/db/records/storage.py
+++ b/apel/db/records/storage.py
@@ -67,9 +67,8 @@ class StorageRecord(Record):
         self._datetime_fields = ["CreateTime", "StartTime", "EndTime"]
         # Fields which will have an integer stored in them
         self._int_fields = ["FileCount", "ResourceCapacityUsed", "LogicalCapacityUsed", "ResourceCapacityAllocated"]
-        
-    
-    def get_apel_db_insert(self, apel_db, source):
+
+    def get_apel_db_insert(self, source=None):
         '''
         Returns record content as a tuple, appending the source of the record 
         (i.e. the sender's DN).  Also returns the appropriate stored procedure.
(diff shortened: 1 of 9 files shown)


test/test_storage_record.py
```diff
@@ -79,6 +79,25 @@ def test_get_apel_db_insert(self):
         record.get_apel_db_insert()
         record.get_apel_db_insert(None)
 
+    def test_get_ur(self):
+        """Check that al input values are present in XML output of get_ur"""
```
gregcorbett: all

apel/db/records/storage.py
```diff
@@ -88,3 +91,121 @@ def get_db_tuple(self, source=None):
         The source (DN of the sender) isn't used in this record type currently.
         """
         return Record.get_db_tuple(self)
+
+    def get_ur(self, withhold_dns=False):
```
gregcorbett: could this be done with a `for` loop? See this similar method for [DataSetSummaryRecord.get_ur](https://github.com/gregcorbett/apel/blob/dataset_accounting_v2/apel/db/records/dataset_summary.py#L76)

test/test_storage_record.py
```diff
@@ -61,6 +62,61 @@ def test_mandatory_fields(self):
...
+
+        XML_HEADER = '<?xml version="1.0" ?>'
+        UR_OPEN = ('<sr:StorageUsageRecords xmlns:sr="http://eu-emi.eu/namespac'
+                   'es/2011/02/storagerecord">')
+        UR_CLOSE = '</sr:StorageUsageRecords>'
+
+        parser = StarParser(XML_HEADER + UR_OPEN + ur + UR_CLOSE)
+
+        # This works as all dictionary values are hashable/immutable
+        missmatched = (set(parser.get_records()[0]._record_content.items()) ^
```
gregcorbett: what does `^` do here? If I read it correctly its an operation on two `set` objects?

apel/db/records/storage.py
tofu-rocketry: Probably. This was a copy-paste job from the manual changes that had been made on the old pre-prod host that I knew worked so didn't redo it.

gregcorbett: okay

test/test_storage_record.py
tofu-rocketry: It's the "symmetric difference". It creates a new set with elements in either the first or second set but not both: https://docs.python.org/2/library/sets.html#set-objects

gregcorbett: ah okay, shiny!

tofu-rocketry:
I've done a little tidy up so there's a few more commits to review.

PR #146, opened 2017-09-13
Add a PID check to the summariser and corresponding unit test 

### LINKED ISSUE #79: Add PID check to summariser
The apel loader and ssm receiver bin scripts check for the existence of a PID file before running and fail if it exists. The summariser bin script does not do this. It is therefore possible to have 2 summarisers running at the same time. We should incorporate the check in the summariser so that this does not happen.

gregcorbett:
Resolves #79 

Add a PID check to the summariser so that two summarisers of the same type can't be run at the same time. If the summariser finds it's Pidfile already exists, or it fails to create a new Pidfile, the summariser will now not run. At then end of the summariser run it tries to clean up the Pidfile and logs if the clean up fails.

Also adds a unit test of `bin/summariser.py` to test this new added functionality. This required a change to `bin/summariser.py` as `runprocess()` previously could not be run without running the python script.

### DIFF
--- a/bin/summariser.py
+++ b/bin/summariser.py
@@ -29,6 +29,7 @@ from apel.db import ApelDb, ApelDbException
 from apel.common import set_up_logging, LOG_BREAK
 from apel import __version__
 
+
 def runprocess(db_config_file, config_file, log_config_file):
     '''Parse the configuration file, connect to the database and run the 
        summarising process.'''
@@ -37,6 +38,9 @@ def runprocess(db_config_file, config_file, log_config_file):
         # Read configuration from file
         cp = ConfigParser.ConfigParser()
         cp.read(config_file)
+
+        pidfile = cp.get('summariser', 'pidfile')
+
         dbcp = ConfigParser.ConfigParser()
         dbcp.read(db_config_file)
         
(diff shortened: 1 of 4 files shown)


gregcorbett:
I have rebased this on top of 1.6.1-1

test/test_summariser.py
```diff
@@ -0,0 +1,111 @@
...
+        os.close(self.sum_cfg)
+
+        # Run the summariser with the temporary config files
+        try:
+            summariser.runprocess(self.db_cfg_path, self.sum_cfg_path, '')
+        except SystemExit:
+            # A SystemExit is raised regardless of what happens
+            # in the summariser, so we must look at the log output
+
+            # Note: Because we need to be compatible with Python 2.4, we can't
```
tofu-rocketry: Can be updated to 2.6

tofu-rocketry: Done.

tofu-rocketry:
Could you have a [look at my changes](https://github.com/apel/apel/pull/146/files/72bc46d85f85fb097b08967827143d874e58d92c..79ba7ecb5d17070492e0de90843e0488f8100f4a), @gregcorbett? I'll probably squish some of the commits once we're both happy before merging this PR.

bin/summariser.py
```diff
@@ -63,8 +63,8 @@ def runprocess(db_config_file, config_file, log_config_file):
 
     # set up logging
     try:
-        if os.path.exists(options.log_config):
-            logging.config.fileConfig(options.log_config)
+        if os.path.exists(log_config_file):
```
tofu-rocketry: Good job spotting this inconsistency.

gregcorbett:
LGTM!

PR #153, opened 2018-03-01
Extra SLURM statuses

tofu-rocketry:
As reported on [GGUS](https://ggus.eu/index.php?mode=ticket_info&ticket_id=126954).

Not enough statuses are being accepted by the SLURM parser, so I've added all the ones that seem to indicate that the job has stopped and resources have been used.

Details on the statuses [here](https://slurm.schedmd.com/sacct.html). It's not the clear what some of them mean (mainly DEADLINE), but it should be a lot better than before.

Tested successfully by reporting user.

### DIFF
--- a/apel/parsers/slurm.py
+++ b/apel/parsers/slurm.py
@@ -98,9 +98,7 @@ class SlurmParser(Parser):
         # Input checking
         if rc['CpuDuration'] < 0:
             raise ValueError('Negative CpuDuration value')
-
-        if rc['WallDuration'] < 0:
-            raise ValueError('Negative WallDuration value')
+        # No negative WallDuration test as parse_time prevents that.
 
         if rc['StopTime'] < rc['StartTime']:
             raise ValueError('StopTime less than StartTime')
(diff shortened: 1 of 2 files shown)


tofu-rocketry:
My first one in a while! Codacy grumbled until I added that docstring.

I might tweak the unit test before finalising this though.

tofu-rocketry:
OK, I'm done faffing with this now.

gregcorbett:
How does `parse_time` protect against negative time stamps of the form `'-0:00:16'`?

I tried running it and got this:
```
>>> from apel.common import parse_time
>>> parse_time('-0:00:16')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "apel/common/datetime_utils.py", line 69, in parse_time
    return 86400*int(days) + 3600*int(hours) + 60*int(minutes) + int(seconds)
ValueError: invalid literal for int() with base 10: ''
```

tofu-rocketry:
I meant in the sense that it errors and doesn't pass them back to the SLURM parser, so the test there is pointless.

`parse_time` uses dashes to split on days and errors if something isn't of the format `[d-]h:m:s`.

PR #167, opened 2018-08-17
Add support for different CPU time fields for SLURM

### LINKED ISSUE #160: Wrong SLURM field used for CpuDuration
From GGUS: [ICM](https://ggus.eu/index.php?mode=ticket_info&ticket_id=132922), [BelGrid-UCL](https://ggus.eu/index.php?mode=ticket_info&ticket_id=132925), and [UKI-LT2-QMUL](https://ggus.eu/index.php?mode=ticket_info&ticket_id=132929)

`slurm_acc.sh` currently uses `CPUTimeRAW` from `sacct`, which is elapsed time * number of CPUs. This results in Elapsed time being the same as the wall clock time.

What should be used is probably `TotalCPU`, which is the sum of the SystemCPU and UserCPU time used by the job. However, parsing this field is less trivial than `CPUTimeRAW` as it seems to use a format which is something like `dd-hh:mm:ss`.

tofu-rocketry:
Resolves #160.

This PR adds support for different time formats of CPU duration to be passed to the SLURM parser. This is to allow for the use of either `CPUTimeRAW` or `TotalCPU` from `sacct` which each have good and bad points depending on use case. I have decided not to enforce either format as the new field TotalCPU may not be so good for multicore jobs. **I've left it up to the user as to which they use.**

I've also added the different CPU time options to the `slurm_acc.sh` script, added unit tests, and tidied some whitespace.

### DIFF
--- a/apel/common/datetime_utils.py
+++ b/apel/common/datetime_utils.py
@@ -12,7 +12,7 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
-   
+
    @author Konrad Jopek, Will Rogers
 '''
 
@@ -25,7 +25,7 @@ def valid_from(date, days=1):
     '''
     Method for BlahParser
     Returns calculated value for ValidFrom field.
-    
+
     By default it returns Timestamp - 1 day
     '''
     delta = datetime.timedelta(days=days)
@@ -36,7 +36,7 @@ def valid_until(date, days=28):
     '''
     Method for BlahParser
     Returns calculated value for ValidUntil field.
-    
+
     By default it returns Timestamp + 28 days
     '''
     delta = datetime.timedelta(days=days)
@@ -47,9 +47,9 @@ def parse_timestamp(datetime_string):
     '''
     Parse timestamp encoded as a string in various forms.  Return
     a TZ-unaware datetime object, which is in UTC.
-    
+
     If timezone information is not present in the string, it
-    assumes that the timezone is UTC.  
+    assumes that the timezone is UTC.
     '''
     dt = iso8601.parse_date(datetime_string)
     utcdt = dt.astimezone(iso8601.iso8601.UTC)
--- a/apel/common/datetime_utils.py
+++ b/apel/common/datetime_utils.py
@@ -58,14 +58,19 @@ def parse_timestamp(datetime_string):
 
 
 def parse_time(timestring):
-    '''
-    Return seconds from times of the form d-h:m:s or h:m:s.
-    '''
+    """Return integer seconds from times of form d-h:m:s, h:m:s or m:s.s."""
     if '-' in timestring:
         days, sub_days = timestring.split('-')
     else:
         days, sub_days = 0, timestring
-    hours, minutes, seconds = sub_days.split(':')
+
+    if '.' in sub_days:
+        hours = 0
+        minutes, decimal_s = sub_days.split(':')
+        seconds = int(round(float(decimal_s)))
+    else:
+        hours, minutes, seconds = sub_days.split(':')
+
     return 86400*int(days) + 3600*int(hours) + 60*int(minutes) + int(seconds)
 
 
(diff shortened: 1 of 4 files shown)


apel/parsers/slurm.py
```diff
@@ -63,6 +63,14 @@ def parse(self, line):
                               'NODE_FAIL', 'PREEMPTED', 'TIMEOUT'):
             return None
 
+        # Select CPU time parsting function based on field used.
```
kjopek: Typo: parsing instead of parsting.

scripts/slurm_acc.sh
```diff
@@ -1,8 +1,17 @@
...
 
+# This script can be run to extract usage accounting data from SLURM for later
+# parsing by an APEL client.
+
 sleep 2
 
 NOW=$(date +"%Y%m%d")
 FILE="/var/log/apel/slurm_acc.$NOW"
 
+# This is the original sacct call which uses CPUTimeRAW for CPU Duration.
```
gregcorbett: Is the intention here sites would use one `sacct` command or the other? If so, maybe a comment stating that would be helpful.

test/test_slurm.py
```diff
@@ -45,6 +45,16 @@ def test_parse_line(self):
             ('297720.batch|batch|||2013-10-25T12:11:20|2013-10-25T12:11:36|00:00:16|16||1|1|wn37|3228K|23820K|COMPLETED'),
             ('321439.batch|batch|||2013-10-27T17:09:35|2013-10-28T04:47:20|11:37:45|41865||1|1|wn16|770728K|1.40G|COMPLETED'),
             ('320816.batch|batch|||2013-10-27T14:56:03|2013-10-28T05:03:50|14:07:47|50867||1|1|wn33|1325232K|2.22G|COMPLETED'),
+            # These lines use TotalCPU rather than CPUTimeRAW.
```
gregcorbett: So these new test cases make use of the `d-h:m:s` syntax as highlighted in bold?

> ('5940815|cream_877733810|user1|group1|2018-06-03T02:11:12|'
>  '2018-06-05T03:36:08|**2-01:24:56**|**8-15:40:36**|partition|8|1|'
>  'wn1|||COMPLETED'),

tofu-rocketry: Kinda. There should be a test for each new format: `d-h:m:s`, `h:m:s` and `m:s.s`

gregcorbett: ah yes, I found it quite hard to parse the test cases - maybe add a comment stating which new test case tests what format?

scripts/slurm_acc.sh
gregcorbett: 👍

test/test_slurm.py
tofu-rocketry: Done.

tofu-rocketry:
I've rebased and squashed the extra commits into the earlier ones. Could I get a review from someone?

apel/common/datetime_utils.py
```diff
@@ -58,14 +58,19 @@ def parse_timestamp(datetime_string):
...
-    Return seconds from times of the form d-h:m:s or h:m:s.
-    '''
+    """Return integer seconds from times of form d-h:m:s, h:m:s or m:s.s."""
     if '-' in timestring:
         days, sub_days = timestring.split('-')
     else:
         days, sub_days = 0, timestring
-    hours, minutes, seconds = sub_days.split(':')
+
+    if '.' in sub_days:
```
jrha: Maybe I just like regexps too much, but I'd probably try to parse the times with something like:
`(?:(?P<days>\d)-)?(?:(?P<hours>\d{2}):)?(?P<minutes>\d{2}):(?P<seconds>\d{2}(?:\.\d+)?)`

Although... maybe you could loop over a series of `strptime()` formats, but there would be problems with day numbering and fractional seconds... nope, regexps FTW.

tofu-rocketry: Feel free to submit a PR! :wink: 

It's a fair point, but I wanted to fix the current problem rather than introduce a new factor in.

PR #170, opened 2018-08-29
Support HTCondorCE

geonmo:
This PR is created to support HTCondorCE accounting info.

Actually, HTCondorCEParser is almost same blahd parser.

- A script "htcondorce_acc.sh" creates accounting information from HTCondorCE using condor_ce_history.
- HTCondorCEParser fills the "BlahdRecord" using HTCondorCE accounting info.
- "apelparser" supports HTCondorCEParser like as BlahdParser.


apel/db/records/htcondorce.py
```diff
@@ -0,0 +1,43 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+'''
+
+from apel.db.records.record import Record
+
+
+class HTCondorCERecord(Record):
```
tofu-rocketry: If this is exactly the same as a blahd record, why not just use that?

scripts/htcondorce_acc.sh
```diff
@@ -0,0 +1,46 @@
...
+NOW=$(date +"%Y%m%dT%H%M%S")
+OUTPUT_FILE=$OUTPUT_LOCATION/accounting.$NOW
+
+# Build the filter for the history command
+CONSTR="(JobStartDate>0)&&(CompletionDate>`date +%s -d "01 Jan 2014"`)"
+
+# Populate the temporary file
+for HF in $HISTORY_FILES
+do
+  $CONDOR_LOCATION/bin/condor_history -file $HF -constraint $CONSTR \
```
tofu-rocketry: Where does `CONDOR_LOCATION` come from?

scripts/htcondorce_acc.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+CONDOR_CE_HISTORY_LOCATION=/bin
```
tofu-rocketry: Is this used?

tofu-rocketry:
Does this support both HTCondor and non-HTCondor batch systems running behind an HTCondor-CE?

geonmo:
@tofu-rocketry I am sorry for late reply. This feature will fill blahp record using condor_ce_history(not condor_history)  command on DB. It means that it can use with other batch systems if it works as intended. However, I did not test for other batch systems. In addition, I did not develop a switch to select blahp or htcondor-ce.  (to prevent to use both.)

geonmo:
@tofu-rocketry I guess that this code is not mature yet. I will make some testing environment. However, this code can help apel user to make their apel client.

sjones-hep-ph-liv-ac-uk:
I've obtained these changes and applied them to a HTCondorCE test framework that I've built up, that uses apel version 1.6.1-1.el7 rpms. I found 7 files that had been changed. 

But I found a problem with the patch that prevents a mysql apel backend from being created. The problem (I think) is that mysql.py tries to make a hash of backends related to MYSQL_TABLES  that uses HTCondorCERecord as a reference. But no class called HTCondorCERecord is imported (or exists in the tree). Hence a NameError exception happens. Can you comment on that? I think this may have occurred when a BlahdRecord was used instead (which has a similar format). Perhaps I'm wrong. Please let me know. We have a few customers lined up for a version of APEL that works with HTCondorCE, including the groups at Madison and WLCG itself.

Please let me know. Ste.

tofu-rocketry:
Closing in favour of #179.

kreczko:
Aren't #170 and #179 completely independent PRs?
This one adds the parsing, the latter just adds an additional parameter.
Or am I misunderstanding something?

sjones-hep-ph-liv-ac-uk:
On 2018-11-30 15:13, Luke Kreczko wrote:
> Aren't #170 [1] and #179 [2] completely independent PRs?

They solve the same problem, in slightly different ways. It turns out 
that we can reuse the standard, existing parsers (BLAH for CE logs, 
htcondor.py for batch logs) if we configure HTCondorCE to write in "BLAH 
format".  And it can do that, with only slight help from (say) bash, sed 
or perl. BTW: We make HTCondorCE wrote in BLAH format by editing the 
condor_history data extraction script (such a script is necessary for 
both #170 and #179.)

So, with #179, we put the onus on the CE to write in the standard 
format, as it is meant to in the general apelclient architecture 
(instead of inventing new BLAH-like formats for every imaginable CE that 
could ever come out.)

> This one adds the parsing, the latter just adds an additional 
> parameter.

All we need to do is add scaling to the htcondor.py parser. That is all 
that was missing. So I took that bit from the PIC solution.

kreczko:
Ah, OK. thank you for the clarification

PR #171, opened 2018-09-20
Split test requirements out and add constraint

tofu-rocketry:
As per apel/ssm#81.

- Move test requirements into separate requirements file to make it
clearer that they are only needed for running unit tests.
- Add constraints file called from the test requirements file to limit
one of coveralls' dependencies to an earlier version that works with
Python 2.6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,9 @@ cache: pip
 # Avoid pip log from affecting cache
 before_cache: rm -fv ~/.cache/pip/log/debug.log
 
-# Install defaults to "pip install -r requirements.txt"
+install:
+  - pip install -r requirements.txt
+  - pip install -r requirements-test.txt
 
 # Commands that prepare things for the test
 before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,12 +17,12 @@ cache: pip
 before_cache: rm -fv ~/.cache/pip/log/debug.log
 
 install:
-  # Base requirements
+  # Base requirements for apel
   - pip install -r requirements.txt
   # Additional requirements for the unit and coverage tests
   - pip install -r requirements-test.txt
 
-# Commands that prepare things for the test
+# Commands to prepare environment for the test
 before_script:
   - export PYTHONPATH=$PYTHONPATH:`pwd -P`
   - cd test
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,9 @@ cache: pip
 before_cache: rm -fv ~/.cache/pip/log/debug.log
 
 install:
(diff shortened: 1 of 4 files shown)


requirements-test.txt
```diff
@@ -0,0 +1,6 @@
+# Constraints on requirements below
```
jrha: I'm not sure if the name or purpose of this file is clear enough, who is supposed to know they should use it?

tofu-rocketry: Do you mean there should be more information here on in the actual constraints file?

tofu-rocketry: Howzat?

.travis.yml
```diff
@@ -16,7 +16,11 @@ cache: pip
 # Avoid pip log from affecting cache
 before_cache: rm -fv ~/.cache/pip/log/debug.log
 
-# Install defaults to "pip install -r requirements.txt"
+install:
+  # Base requirements
+  - pip install -r requirements.txt
+  # Additional requirements for the unit and coverage tests
+  - pip install -r requirements-test.txt
```
gregcorbett: Does this work due to the way `-r` handles `requirements-test.txt`?
Could we have `pip install -r requirements-test.txt -c constraints.txt` in the install section of .travis.yml instead?

tofu-rocketry: Yes, it does (for a start the build passes).

The point of linking the constraints file from inside the requirements file is to keep everything in one place and ensure that it isn't missed. Also, you only need to look at the requirements file and not the Travis file as well to figure out where it's used.

gregcorbett: fair enough

requirements-test.txt
tofu-rocketry: So, having had a think and done a bit of a research, I've decided to keep the filename as `requirements-test.txt` is quite a common name for this file and also it matches the pluralisation of the directory for the unittests (which is also a common name for that). I will try and improve the flow of the comments though.

constraints.txt
```diff
@@ -0,0 +1,4 @@
+# Constraints that apply to requirements-test.txt
```
gregcorbett: If we're documenting that `constraints.txt` (only) applies to `requirements-test.txt`, should we call it `constraints-test.txt`?

tofu-rocketry: If we need constraints on the base requirements in the future, I'd rather use the same file, so I'll change the comment to be less specific.

gregcorbett:
I am happy to approve the changes, do you want to squash the commits down before I review though?

tofu-rocketry:
I can do the squash through the GH interface which preserves the approval.

PR #174, opened 2018-10-17
Use wildcard glob matching to ensure setup.py installs any update scripts under scripts, to match apel.spec 

### LINKED ISSUE #165: Install via setup.py fails, wants to install scripts/update_schema.sql
[`setup.py`](https://github.com/apel/apel/blob/dev/setup.py#L59) wants to install `scripts/update_schema.sql`

But `scripts/update_schema.sql` doesn't exist as it was renamed to include version numbers (and now there are multiple versions of it.

This causes `python setup.py install` to fail.

Possible solutions:
- Wild card it like we did in the spec file.
- Don't include the update scripts in an install via `setup.py`.
- Have a post test step in TravisCI that tries to install via `setup.py`.

gregcorbett:
Resolves #165 

This prevents having to manually update the list of update scripts so long as the current naming convention continues.

I also took this opportunity to move the directory creation (`/var/log/apel` and `/var/run/apel`) into the `setup` call, rather than as hacky `if` statements.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,7 @@ Requires setuptools.
 """
 
 import glob
-from os import remove, path
+from os import remove
 from shutil import copyfile
 import sys
 
--- a/setup.py
+++ b/setup.py
@@ -94,7 +94,8 @@ def main():
                       (log_rotate_dir, log_rotate_files),
                       # Create empty directories
                       ('/var/log/apel', []),
-                      ('/var/run/apel', [])],
+                      ('/var/run/apel', [])
+                      ],
           # zip_safe allows setuptools to install the project
           # as a zipfile, for maximum performance!
           # We have disabled this feature so installing via the setup
--- a/setup.py
+++ b/setup.py
@@ -11,6 +11,7 @@ Usage: 'python setup.py install'
 Requires setuptools.
 """
 
+import glob
 from os import remove, path
 from shutil import copyfile
 import sys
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -94,7 +91,10 @@ def main():
                       (data_dir, accounting_files),
                       (data_dir, message_files),
                       (data_dir, update_scripts),
-                      (log_rotate_dir, log_rotate_files)],
+                      (log_rotate_dir, log_rotate_files),
+                      # Create empty directories
+                      ('/var/log/apel', []),
+                      ('/var/run/apel', [])],
```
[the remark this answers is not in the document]
gregcorbett: `pylint` and `pep8` seem to conflict as to how many spaces this bracket should be indented. I'm going to go with `pep8` so codeclimate doesn't complain.

gregcorbett:
apologies. Change applied and rebased.

setup.py
```diff
@@ -94,7 +91,11 @@ def main():
                       (data_dir, accounting_files),
                       (data_dir, message_files),
                       (data_dir, update_scripts),
-                      (log_rotate_dir, log_rotate_files)],
+                      (log_rotate_dir, log_rotate_files),
+                      # Create empty directories
+                      ('/var/log/apel', []),
+                      ('/var/run/apel', [])
```
jrha: You should be able to add a trailing `,` at the end of this line, Python will ignore and you will be able to cleanly add more entries later.

tofu-rocketry: Oh, yes. I forgot to add this in my suggestion.

gregcorbett: done

setup.py
```diff
@@ -11,7 +11,8 @@
 Requires setuptools.
 """
 
-from os import remove, path, makedirs
+import glob
+from os import remove, path
```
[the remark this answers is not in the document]
gregcorbett: removed

PR #175, opened 2018-10-30
Clarify what local_jobs and send_ur options do in unloader config

gregcorbett:
specifically, clarify that `send_ur` is also used for storage accounting. As someone setting up a regional APEL server for storage accounting might not know that `send_ur` has to be true to unload storage records.

### DIFF
--- a/conf/unloader.cfg
+++ b/conf/unloader.cfg
@@ -13,8 +13,12 @@ dir_location=/var/spool/apel/
 # VStarRecords
 table_name = VJobRecords
 
-# Only for job records and summaries
+# Only used when sending job records, job summaries or storage records.
+# - for grid accounting, if send_ur is true, records and summaries will be sent
+#   in the OGF Usage Record format.
+# - for storage accounting, send_ur must be true.
 send_ur = false
+# Only for job records and summaries, to report on local jobs at the site.
 local_jobs = true
 
 withhold_dns = false


conf/unloader.cfg
```diff
@@ -13,8 +13,12 @@ dir_location=/var/spool/apel/
 # VStarRecords
 table_name = VJobRecords
 
-# Only for job records and summaries
+# Only used for job and storage accounting.
```
tofu-rocketry: Is this option useable for summary records? The change makes that less clear.

tofu-rocketry: `send_ur` MUST be set to true for storage, right? That should be made clear, otherwise this change doesn't address your main point the the PR description.

gregcorbett: My reasoning was that its used for job accounting regardless of whether you are sending summaries or individual job records.

I could change it to:
`#  Only used when sending job records, job summaries or storage records.`

gregcorbett: how about I change that bit to something like
```
# For job accounting, if send_ur is true, records and summaries will be sent in the OGF Usage Record format.
# For storage accounting, send_ur must be true.
```

tofu-rocketry: Yes, that's good.

tofu-rocketry: Yes, that's better. Just change "job" to "grid".

tofu-rocketry:
Could you please rebase while you're at it?

gregcorbett:
Changed and rebased

PR #178, opened 2018-11-29
Add cputmult to htcondor.py. Also new test case.

sjones-hep-ph-liv-ac-uk:
The new, optional, cputmult factor will, if present in the accounting line, apply a 
scaling factor to wall abd cpu durations. If not present, the scaling factor defaults to 1.0, i.e. no 
change. This will allow apelclient to be used in htcondorce deployments.
The unit test for htcondor.py has been expanded to include a run
with this feature. Existing sites who use (say) CREAM/HTCondor (who would not 
include this new field) will be unaffected. The scheme described here:

https://twiki.cern.ch/twiki/bin/view/LCG/HTCondorAccounting


sjones-hep-ph-liv-ac-uk:
Some syntax issue. Will fix.

PR #179, opened 2018-11-29
New, optional, cputmult factor 

sjones-hep-ph-liv-ac-uk:
The new, optional, cputmult factor will, if present in the accounting line, apply a
scaling factor to wall and cpu durations. If not present, the scaling factor defaults to 1.0, i.e. no
change. This will allow apelclient to be used in htcondorce deployments.
The unit test for htcondor.py has been expanded to include a run
with this feature. Existing sites who use (say) CREAM/HTCondor (who would not
include this new field) will be unaffected. The scheme described here:

https://twiki.cern.ch/twiki/bin/view/LCG/HTCondorAccounting

### DIFF
--- a/apel/parsers/htcondor.py
+++ b/apel/parsers/htcondor.py
@@ -43,7 +43,7 @@ class HTCondorParser(Parser):
 
         values = line.strip().split('|')
         cputmult = float(1.0)
-        if (len(values) > 10):
+        if len(values) > 10:
             cputmult = float(values[10])
 
         mapping = {'Site'            : lambda x: self.site_name,
--- a/apel/parsers/htcondor.py
+++ b/apel/parsers/htcondor.py
@@ -53,7 +53,7 @@ class HTCondorParser(Parser):
                    'LocalUserID'     : lambda x: x[1],
                    'LocalUserGroup'  : lambda x: "",
                    'WallDuration'    : lambda x: int(x[2]) * cputmult,
-                   'CpuDuration'     : lambda x: ((int(x[3])+int(x[4])) 
+                   'CpuDuration'     : lambda x: ((int(x[3])+int(x[4]))
                                                   * cputmult),
                    'StartTime'       : lambda x: x[5],
                    'StopTime'        : lambda x: x[6],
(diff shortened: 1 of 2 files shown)


tofu-rocketry:
Thanks. I'll have a look at it after the GridPP Technical Meeting.

jrha:
@tofu-rocketry regarding the discussion earlier today, this concept is called "host_factor" in the LSF parser and `cputmult`/`wallmult` in the SGE parser, so is already inconsistent!

sjones-hep-ph-liv-ac-uk:
The variable name, cputmult, is only relevant for reading the code. The name "cputmult" is not "exposed" in the htcondor.py parser; the fields in the data do not have "names", merely an index. Field order, therefore, is important, but not what the field is called in the code (or the documentation.) It can be anything, and it would still work.

tofu-rocketry:
Don't worry @sjones-hep-ph-liv-ac-uk - James' point is mainly about internal consistency within the APEL codebase. It's something we (APEL) can change later if it we find it confusing.

I have a few commits to add on top of this and then I think we'll be ready to merge. It would be good if you could email us a copy of the Condor side scripts you used with this and then we can think about where to put them.

sjones-hep-ph-liv-ac-uk:
The condor side scripts used with this are only test versions, and not portable yet. So the next step is to rig up a real HTCondor-CE system (I've been using stubs until now), develop acceptable extraction scripts, and do an end-to-end system test. This is in progress; deadline is Feb, but I expect results before end of year. Then I'll deliver the scripts. BTW: there is no problem with releasing in stages; this change is unit tested and compatible with existing use cases.

tofu-rocketry:
Ok, I'll go ahead and finish this one off then.

tofu-rocketry:
I should probably correct commit author before merging.

geonmo:
@sjones-hep-ph-liv-ac-uk Is there a part that uses "condor_ce_history"? I think you can make more clearly if you use "condor_ce_history" and "condor_history". As you can see in my old code, you can use the "GlobalJobId" and "RoutedJobId" from condor_ce_history to match the local condor job from condor_history.

sjones-hep-ph-liv-ac-uk:
> Is there a part that uses "condor_ce_history"?

Thanks for this info. We're currently implementing an example "data 
extraction script". The intention is to use a similar approach to PIC, 
but outputting separate files for BLAH format (for  CE log data) and 
standard hycondor.py format (for the batch log data). We'd like to use 
one tool/one pass, but if necessary we'll use "condor_ce_history" for 
one part, and "condor_history" for the other.

tofu-rocketry:
I've run `filter-branch` to correct the email addresses and rebased on the latest work on `dev`.

PR #180, opened 2018-12-05
Add warnings in log if no records unloaded

tofu-rocketry:
We often get people raising tickets when they realise their accounting data isn't displaying and it turns out to be a publishing issue. Sometimes they've not checked their log, and other times they just haven't spotted any problems in the log.

This PR addresses the second case by adding warnings in log if no usage or sync records are unloaded to hopefully make it clearer when something has not gone right.

### DIFF
--- a/bin/client.py
+++ b/bin/client.py
@@ -272,7 +272,10 @@ def run_client(ccp):
                 log.warn('Unrecognised interval: %s', interval)
                 log.warn('Will not start unloader.')
 
-            log.info('Unloaded %d records in %d messages.', recs, msgs)
+            if recs > 0:
+                log.info('Unloaded %d records in %d messages.', recs, msgs)
+            else:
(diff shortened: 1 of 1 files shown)


tofu-rocketry:
It would _usually_ be an issue. If they're sending job records, they would have to have had no jobs finished since the last run of the APEL client, but there would still be sync records. If they're sending summary records, then that would be a problem as they should be sending updated summaries for the last two months on each run of the client.

I guess I could make it more conservative and only warn if there were no sync records, or modify the message for the first case to just prompt the user to check that there were no new jobs run.

jrha:
Or just change the message a bit? You don't want to exchange one form of spurious ticket for another.

Perhaps:
  "No sync records unloaded. If this is unexpected, please check your config."

tofu-rocketry:
There we go. I've also clarified what general type of records are meant in the first warning message.

PR #182, opened 2018-12-07
Update requirements-test.txt

tofu-rocketry:
Restrict coveralls to version that supports Python 2.6


tofu-rocketry:
Probably because the coverage results actually uploaded properly at last. I think there may be minor coverage differences between Python 2.6 and 2.7.

PR #184, opened 2019-01-07
Release 1.8.0 to master

tofu-rocketry:


tofu-rocketry:
This should all be reviewed code, but GitHub only seems to spot that some of the time. :/

Issue #193, opened 2019-06-20
Can't call apel.db.backends.mysql.get_records() for CloudRecords


gregcorbett:
I have created a [test case](https://github.com/gregcorbett/apel/blob/republish_fix_test/test/test_cloud_republish.py#L62) that illustrates the issue.

[Travis](https://travis-ci.org/gregcorbett/apel/jobs/548270794) error message:
```
======================================================================
ERROR: test_cloud_republish (test_cloud_republish.TestRepublish)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test_cloud_republish.py", line 62, in test_cloud_republish
    database.get_records(apel.db.records.cloud.CloudRecord)
  File "/home/travis/build/gregcorbett/apel/apel/db/backends/mysql.py", line 189, in get_records
    for batch in self._get_records(record_type, select_query, rec_number):
  File "/home/travis/build/gregcorbett/apel/apel/db/backends/mysql.py", line 226, in _get_records
    record.set_all(row)
  File "/home/travis/build/gregcorbett/apel/apel/db/records/record.py", line 91, in set_all
    raise InvalidRecordException('Unknown field: %s' % key)
apel.db.records.record.InvalidRecordException: Unknown field: MeasurementMonth
```

Given that the `Measurement{Month, Year, Time}` are database only fields, I'm not sure the best way to fix this...

tofu-rocketry:
I think the issue is that the measurement time fields ([`MeasurementTime, MeasurementMonth, MeasurementYear`](https://github.com/apel/apel/blob/9476bd86424f6162c3b87b6daf6b4270ceb8fea6/schemas/cloud.sql#L518)) aren't specific in the [Python `CloudRecord` definition](https://github.com/apel/apel/blob/dev/apel/db/records/cloud.py#L32).

gregcorbett:
We reckon putting the variables in self._db_fields is sufficient, can't remember off the top of our heads if the order of that list matters.

gregcorbett:
So putting them in `self._db_fields` like this [1] caused a problem [2] because the Measurement values are then passed to the MySQL `ReplaceCloudRecord` function (which we don't want them to be) and then Python rightly complains that not all arguments are converted during string formatting.

Treating them in the same fashion as `UpdateTime` [3], which kinda makes sense because `UpdateTime` and the Measurement values are created in the database, allows records to be loaded and unloaded, but this means the Measurement values don't get unloaded so doesn't help writing a test for #194.

:confounded:

[...]

PR #194, opened 2019-06-20
 Keep the newest cloud record per VM per month 

gregcorbett:
This is a halfway house between the pre-LRVM record loading and the LRVM record loading, where we keep multiple records for a VM (one per month) but take the newest record we receive per month (as opposed to doing some `MeasurementTime` base check).

This makes it easier for sites to republish because they don't need to worry about the state of the repository.

I have created #193 about the problems creating a test case around this.


schemas/cloud.sql
```diff
@@ -119,7 +119,7 @@ BEGIN
         END IF;
     END IF;
 
-    INSERT INTO CloudRecords(RecordCreateTime, VMUUID, SiteID, CloudComputeServiceID, MachineName,
+    REPLACE INTO CloudRecords(RecordCreateTime, VMUUID, SiteID, CloudComputeServiceID, MachineName,
```
tofu-rocketry: What is the benefit of using REPLACE here?

schemas/cloud.sql
```diff
@@ -132,53 +132,6 @@ BEGIN
         suspendDuration, wallDuration, cpuDuration, cpuCount,
         networkType, networkInbound, networkOutbound, publicIPCount, memory, disk,
         benchmarkType, benchmark, storageRecordId, imageId, cloudType, DNLookup(publisherDN))
-      ON DUPLICATE KEY UPDATE
```
tofu-rocketry: Nice to get this simplified.

schemas/cloud.sql
gregcorbett: The benefit is we don't have to think about how the "upsert" is handled, as we currently do with the `INSERT... ON DUPLICATE KEY UPDATE`, `REPLACE` just keeps the last one.

We could achieve the same with `INSERT... ON DUPLICATE KEY UPDATE`, but it wouldn't reduce the complexity of the `ON DUPLICATE KEY UPDATE` very much.

tofu-rocketry: Ah, so we were limited to using `INSERT... ON DUPLICATE KEY UPDATE` before as we needed to do fiddly things with MeasurementTime, right? But now we're not limited to that so `REPLACE` is the simpler option?

gregcorbett: 👍

tofu-rocketry:
I've checked that the update script seems to create the same procedure definition as the base cloud schema.

tofu-rocketry:
I've renamed the update script as the numbers it refers to should bridge the whole gap from the last time the schema was changed. Could of other minor changes.

PR #201, opened 2019-07-31
Allow records per message to be configured in the DB unloader

### LINKED ISSUE #187: Enable a config parameter to set number of records in a message
While investigating increasing the number of records per message that's sent to the Portal, it turned out that the `records` parameter in the `db.cfg` is not actually being used, neither does there seem to be a way to set the number of records returned from the database. The method that does this is `get_records` in `db/backends/mysql.py`.

`db.cfg` may not be the best place to set this. `unloader.cfg` could be more suitable.

Possible impact of change: where users have had this parameter set at a different value, they may start sending messages of that size without realising.

jrha:
Resolves #187.

### DIFF
--- a/bin/dbunloader.py
+++ b/bin/dbunloader.py
@@ -118,17 +118,25 @@ if __name__ == '__main__':
 
     interval = cp.get('unloader', 'interval')
 
-    records_per_message = 1000
+    unloader = DbUnloader(db, unload_dir, include_vos, exclude_vos, local_jobs, withhold_dns)
     try:
-        rpm = int(cp.get('unloader', 'records_per_message'))
-        if rpm < RECORDS_PER_MESSAGE_MIN:
-            records_per_message = RECORDS_PER_MESSAGE_MIN
-            log.warn('records_per_message too small, increasing from %d to %d', rpm, RECORDS_PER_MESSAGE_MIN)
-        elif rpm > RECORDS_PER_MESSAGE_MAX:
-            records_per_message = RECORDS_PER_MESSAGE_MAX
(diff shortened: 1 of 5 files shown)


bin/dbunloader.py
```diff
@@ -114,6 +119,24 @@
...
+        elif rpm > RECORDS_PER_MESSAGE_MAX:
+            unloader.records_per_message = RECORDS_PER_MESSAGE_MAX
+            log.warn('records_per_message too large, decreasing from %d to %d', rpm, RECORDS_PER_MESSAGE_MAX)
+        else:
+            unloader.records_per_message = records_per_message
+    except ConfigParser.NoOptionError:
+        log.info('records_per_message not specified, defaulting to %d.', RECORDS_PER_MESSAGE_DEFAULT)
+        unloader.records_per_message = RECORDS_PER_MESSAGE_DEFAULT
+    except ValueError:
+        log.error('Invalid records_per_message value, must be a postive integer. Defaulting to 1000.')
```
tofu-rocketry: Should insert the default value like you do for the `NoOptionError`.

bin/dbunloader.py
```diff
@@ -114,6 +119,24 @@
     interval = cp.get('unloader', 'interval')
 
     unloader = DbUnloader(db, unload_dir, include_vos, exclude_vos, local_jobs, withhold_dns)
+
+    try:
```
tofu-rocketry: Should should probably factored out to a function that we can test.

jrha: It's just consistent with the other option parsing code.
Happy to open a separate PR that factors all of them out though.

tofu-rocketry: Fair enough.

PR #200, opened 2019-07-31
Add Measurement variables to cloud record ignored fields

gregcorbett:
Resolves #193 

As discussed in that issue, treat `Measurement{Time, Month, Year}` the same as `UpdateTime`, as they are all created in the database on load but we dont necessarily want them being unloaded.

PR #203, opened 2019-08-01
Remove trailing whitespace

jrha:


jrha:
The two "new" issues flagged by codeclimate are pre-existing issues that have been exposed by this whitespace only change.

jrha:
Rebased to fix conflict with #200.

PR #208, opened 2019-12-09
pycodestyle: Only warn about lines longer than 120 characters

jrha:
pycodestyle was formerly pep8 and is what code climate uses.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[pycodestyle]
+max-line-length = 120


gregcorbett:
If we want to extend the max line limit - I'd prefer a slightly lower limit, 110 for example, as 120 still wraps when displaying things side by side for me.

tofu-rocketry:
Sorry, I'm going to stick to our conservative limits for now.

jrha:
:disappointed:

PR #214, opened 2020-03-03
Allow multiple types of service when fetching DNs

### LINKED ISSUE #195: Allow multiple types of service when fetching DNs
See also #77 which could probably be tackled at the same time.

It would be good if `retrieve_dns.py` could accept records from Vac, Vcycle, ARC, HTCondorCE and similar services which don't use the APEL client that are in GOCDB rather than have to register a virtual site APEL service. This would mean checking that the service has a service type from a list rather than just being the site APEL service’s service type.

tswinb:
Resolves #195

### DIFF
--- a/bin/retrieve_dns.py
+++ b/bin/retrieve_dns.py
@@ -33,6 +33,7 @@ import os
 import sys
 import time
 import urllib
+import re
 import xml.dom.minidom
 import xml.parsers.expat
 
@@ -61,6 +62,13 @@ def get_config(config_file):
         c.gocdb_url = cp.get('auth', 'gocdb_url')
     except ConfigParser.NoOptionError:
         c.gocdb_url = None
+
+    try:
+        c.service_types = cp.get('auth', 'service_types')
+    except ConfigParser.NoOptionError:
+        c.service_types = None
+    if c.service_types and not verify_service_types(c.service_types):
+        raise ValueError("Invalid characters in service type.")
 
     try:
         extra_dns = cp.get('auth', 'extra-dns')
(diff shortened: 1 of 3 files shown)


bin/retrieve_dns.py
```diff
@@ -208,6 +216,20 @@ def verify_dn(dn):
 
     return True
 
+def verify_service_types(service_types):
```
gregcorbett:
```suggestion
def verify_service_types(service_types):
    '''
    Returns true if all the provided service types are valid GOCDB service types.
    
    Else, returns false. 
    Validation is done using the regex GOCDB 5.7.4 uses for validation 
    '''
```

bin/retrieve_dns.py
```diff
@@ -223,41 +245,41 @@ def runprocess(config_file, log_config_file):
...
-        log.info(e)
-        fetch_failed = True
-
-    if fetch_failed and (time.time() - os.path.getmtime(cfg.dn_file) <
-                         (cfg.expire_hours * 3600)):
-        log.warn('Failed to update DNs from GOCDB. Will not modify DNs file.')
-        log.info("auth will exit.")
-        log.info(LOG_BREAK)
-        sys.exit(1)
+    for next_url in generate_gocdb_urls(cfg.gocdb_url,cfg.service_types):
```
gregcorbett: As discussed in the hackathon, what happens here if `cfg.gocdb_url` contains a service type because a user hasn't updated their config?

test/test_retrieve_dns.py
```diff
@@ -29,6 +29,22 @@ def test_verify_invalid_dns(self):
             self.assertFalse(bin.retrieve_dns.verify_dn(dn),
                              "Invalid DN accepted: %s" % dn)
 
+    def test_verify_valid_service_types(self):
+        valid_service_types = ("gLite-APEL","uk.ac.gridpp.vcycle",
+                               "1.1234","_test")
+
+        for service_type in valid_service_types:
```
gregcorbett: is it right<sup>TM</sup> to pass each service type in turn to `bin.retrieve_dns.verify_service_types`? It looks like that method "expects" (or atleast can handle) a comma seperated list.

test/test_retrieve_dns.py
```diff
@@ -29,6 +29,22 @@ def test_verify_invalid_dns(self):
...
+            self.assertTrue(bin.retrieve_dns.verify_service_types(service_type),
+                            "Valid service type rejected: %s" % service_type)
+
+    def test_verify_invalid_service_types(self):
```
gregcorbett: see [above comment](https://github.com/apel/apel/pull/214/files#r397799063).

jrha:
Rebased on `dev` and fixed logging conflicts.

gregcorbett (changes requested):
See previous comments.

PR #223, opened 2020-10-08
Add schema version table and version checking code

### LINKED ISSUE #188: Add schema version tables
Add schema version tables to each schema that can be updated when schema update scripts are applied (as well as being initialised with the version number on first creation). Should have a column for when the version was updated.

The could also be queried and printed in the log.

ineilson:
Closes #188


tofu-rocketry:
Reopened as #303.

PR #224, opened 2020-12-07
Remove support for Oracle as a backend database

jrha:
This has not really been supported for a long time, nothing tests it and no
sites are known to be making use of it.


gregcorbett (approved):
Looks good to me!

PR #226, opened 2020-12-09
Add support for loading IP Accounting Messages/Records into a DB

gregcorbett:
Hi all,

This is what I've got so far in terms of adding accounting for Public IPs in a cloud system as part of wider Cloud Accounting in APEL.

Let me know you're thoughts. At the moment [Travis](https://travis-ci.org/github/gregcorbett/apel/jobs/748157851) is complaining because of a dependency issue, it seems a solvable issue as it works on a VM with older versions of dependencies,

This [commit](https://github.com/apel/apel/commit/2ecbd80081dd1fcc9aed0160dc6398c330d53f27) can / should be factored out at.

### DIFF
--- a/apel/common/message_schemas.py
+++ b/apel/common/message_schemas.py
@@ -1,0 +1,149 @@
+"""A file to store message schemas for JSON based messages"""
+
+IP_MSG_SCHEMA = {
+    "definitions": {
+
+    },
+    "$schema": "http: //json-schema.org/draft-07/schema#",
+    "$id": "http: //example.com/root.json",
+    "type": "object",
+    "title": "The Root Schema",
+    "required": [],
+    "properties": {
+        "Type": {
+            "$id": "#/properties/Type",
+            "type": "string",
+            "title": "The Type Schema",
+            "const": "APEL Public IP message"
+        },
+        "Version": {
+            "$id": "#/properties/Version",
+            "type": "string",
+            "title": "The Version Schema",
+            "enum": [
+                "v0.2"
+            ],
+
+        },
+        "UsageRecords": {
+            "$id": "#/properties/UsageRecords",
+            "type": "array",
+            "minItems": 1,
+            "maxItems": 1000,
+            "title": "The UsageRecords Schema",
+            "items": {
+                "$id": "#/properties/UsageRecords/items",
+                "type": "object",
+                "title": "The UsageRecord Schema",
+                "required": [
+                    "MeasurementTime",
+                    "SiteName",
+                    "CloudType",
+                    "LocalUser",
+                    "LocalGroup",
+                    "GlobalUserName",
+                    "FQAN",
+                    "IPVersion",
+                    "IPCount"
+                ],
+                "properties": {
+                    "MeasurementTime": {
+                        "$id": "#/properties/UsageRecords/items/properties/MeasurementTime",
+                        "type": "integer",
+                        "title": "The MeasurementTime Schema",
+                        "examples": [
+                            1536070129
+                        ]
+                    },
+                    "SiteName": {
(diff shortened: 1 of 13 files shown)


apel/common/message_schemas.py
```diff
@@ -0,0 +1,149 @@
+"""A file to store message schemas for JSON based messages"""
+
+IP_MSG_SCHEMA = {
```
gregcorbett, line 3: As a note to Future Greg, we probably want to version this contant in some sense, i.e. `IP_MSG_SCHEMA_V0.2`.

schemas/cloud-extra-public-ip.sql
```diff
@@ -0,0 +1,102 @@
...
+  VOGroupID INT NOT NULL,
+  VORoleID INT NOT NULL,
+
+  IPVersion TINYINT NOT NULL,
+  IPCount INT NOT NULL,
+
+  PublisherDNID INT NOT NULL,
+
+  -- We want to keep one record per site, per cloud, per user, per IP version,
+  -- per month/year.
```
gregcorbett, line 30: As the UsageRecord 'should be a “snapshot” of the number of IPs currently assigned to a user' [[1](https://egi-federated-cloud.readthedocs.io/en/latest/federation.html?highlight=public%20ip%20accounting#public-ip-usage-record)], maybe we should keep one record per day and build a monthly summary out of those daily records. 

Otherwise, what would we do if a User had 1 IP on Jan 1st, and 100 IPs on Jan 31st, and 0 in between?

Granted that case can only really be solved by using the timestamp in the key... otherwise you can just say "what would we do if a User had 1 IP at 9am on Jan 1st, and 100 IPs at 11pm on Jan 1st, and 0 in between?". Maybe one for the FedCloud to weigh in on.

gregcorbett: Reporting a combination of minimum, maximum and average number of IPs assigned to a user in a given accounting period (month) seems like the way forward, which probably means keeping 1 record per day in order to calculate those values.

apel/common/message_schemas.py
```diff
@@ -0,0 +1,149 @@
...
+    "definitions": {
+
+    },
+    "$schema": "http: //json-schema.org/draft-07/schema#",
+    "$id": "http: //example.com/root.json",
+    "type": "object",
+    "title": "The Root Schema",
+    "required": [],
```
gregcorbett, line 11: We should probably explicitly express that these fields are required, as if any of them are missing the message wont/shouldn't be parsed properly.

```suggestion
    "required": [
        "Type",
        "Version",
        "UsageRecords"
    ],
```

apel/common/message_schemas.py
```diff
@@ -0,0 +1,149 @@
...
+                    "MeasurementTime",
+                    "SiteName",
+                    "CloudType",
+                    "LocalUser",
+                    "LocalGroup",
+                    "GlobalUserName",
+                    "FQAN",
+                    "IPVersion",
+                    "IPCount"
+                ],
```
gregcorbett, line 48: Where fields (like `LocalUser` can be `"null"` do we require them to be in the message?

apel/common/message_schemas.py
```diff
@@ -0,0 +1,149 @@
...
+                    },
+                    "GlobalUserName": {
+                        "$id": "#/properties/UsageRecords/items/properties/GlobalUserName",
+                        "type": "string",
+                        "title": "The GlobalUserName Schema",
+                        "examples": [
+                            "GlobalUser1"
+                        ],
+
+                    },
```
gregcorbett, line 115: GlobalUserName should be nullable because the tenancy model in OpenStack is based on projects and there will be IPs allocated to a project, not owned by any user.

PR #229, opened 2021-01-06
 Adds summarising procedure for storage accounting

### LINKED ISSUE #198: Storage: No summaries or LastUpdated table
Need to add method to summarise storage records.

As part of this we may need to add a `LastUpdated` table as per the other types of accouting.

sarahbyrnie:
Resolves #198


bin/summariser.py
```diff
@@ -109,6 +109,7 @@ def runprocess(db_config_file, config_file, log_config_file):
             # Make sure that records are not coming from the same site by two different routes
             # N.B. Disabled as check doesn't scale well and isn't that useful.
             # db.check_duplicate_sites()
+            db.summarise_storage()
```
gregcorbett, line 112: As discussed, this should be moved into it's own `if db_type == 'storage':` block.

schemas/storage.sql
```diff
@@ -125,7 +125,7 @@ CREATE TABLE UserIdentities (
     INDEX(NAME)
 );
 
-DROP FUNCTION IF EXISTS UserIdentityLookup;
+DROP FUNCTION IF EXISTS UserIdentitiesLookup;
```
gregcorbett, line 128: Why change this? Is this intended to change the name of the function `UserIdentityLookup` to `UserIdentitiesLookup`?

If so
- Shouldn't the subsequent `CREATE FUNCTION UserIdentityLookup` should also be changed.
- Presumably we'd need an update schema script similar to https://github.com/apel/apel/blob/dev/scripts/update-1.7.0-1.8.2.sql

gregcorbett (changes requested):

schemas/storage.sql
```diff
@@ -374,3 +374,51 @@ WHERE StarRecords.StorageSystemID = StorageSystems.id
...
+          DirectoryPath,
+          `Group`,
+          EndTime,
+          MONTH(EndTime),
+          YEAR(EndTime),
+          SUM(FileCount),
+          SUM(ResourceCapacityUsed),
+          SUM(LogicalCapacityUsed),
+          SUM(ResourceCapacityAllocated)
+   FROM VStarRecords
```
gregcorbett, line 422:
```suggestion
   FROM StarRecords
```

Should this run against the underlying table? not the view?

schemas/storage.sql
```diff
@@ -374,3 +374,51 @@ WHERE StarRecords.StorageSystemID = StorageSystems.id
   AND StarRecords.GroupID = Groups.id
   AND StarRecords.SubGroupID = SubGroups.id
   AND StarRecords.RoleID = Roles.id;
+
+-- -----------------------------------------------------------------------------
+-- Summaries
+DROP TABLE IF EXISTS DaySummaries;
+CREATE TABLE DaySummaries (
+  Site                        VARCHAR(255),
```
gregcorbett, line 382: See https://github.com/apel/apel/pull/229/files#r1188726030, if so - should all the VARCHAR rows be INTs instead? Like this:

```suggestion
  SiteID                        INT NOT NULL),
```

schemas/storage.sql
```diff
@@ -374,3 +374,51 @@ WHERE StarRecords.StorageSystemID = StorageSystems.id
...
+          `Group`,
+          EndTime,
+          MONTH(EndTime),
+          YEAR(EndTime),
+          SUM(FileCount),
+          SUM(ResourceCapacityUsed),
+          SUM(LogicalCapacityUsed),
+          SUM(ResourceCapacityAllocated)
+   FROM VStarRecords
+   Group BY 1,2,3,4,5,6,7;
```
gregcorbett, line 423: Do we want to group by EndTime? If two records differed only by a second in the endtime we'd get two summaries. Would Month(EndTime) and Year(EndTime) be better?

schemas/storage.sql
gregcorbett: @Sae126V, can you add additional commits to the PR to change this (and https://github.com/apel/apel/pull/229/files#r1188726030)? The procedures this PR has added will likely need to be updated.

There are already lookup functions in the storage schema which should handle the lookup of a names to IDs (i.e. `Site` to `SiteID`).

This might have some knock on effects for https://github.com/apel/apel/pull/315.

schemas/storage.sql
tofu-rocketry: Yes, it should be by month and year, for example how we do it with grid: https://github.com/apel/apel/blob/0f888108a90298ab2bdc3ed0a3d0c32575ab84ea/schemas/server.sql#L293

Also, the `group by` should be by explicitly named columns. Using numbered columns is fine on the command line, but we should write it out in full for a schema in the repository.

schemas/storage.sql
tofu-rocketry: Yes. Will need to check which columns should be using the "ID" reference table version of the field

PR #233, opened 2021-03-11
Refactoring SSM call to make client AMS compatible

### LINKED ISSUE #230: Client doesn't support using SSM with AMS
As reported [here ](https://ggus.eu/index.php?mode=ticket_info&ticket_id=150029)and [here](https://ggus.eu/index.php?mode=ticket_info&ticket_id=148981)

The `run_ssm` function doesn't support invoking SSM for AMS sending: https://github.com/apel/apel/blob/6193531424f4204349c02ae062a1586fe46b4cc7/bin/client.py#L58

tofu-rocketry:
Resolves #230.

Depends on apel/ssm#149.

### DIFF
--- a/apel.spec
+++ b/apel.spec
@@ -44,7 +44,7 @@ supported by the APEL system: Torque, SGE and LSF.
 %package client
 Summary:        APEL client package
 Group:          Development/Languages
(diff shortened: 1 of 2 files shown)


gregcorbett (approved):
LGTM :)

PR #234, opened 2021-03-19
Update dependencies

### LINKED ISSUE #211: Add python-dirq and check others dependencies
Issue raised in GGUS [here](https://ggus.eu/index.php?mode=ticket_info&ticket_id=143490).

Changes to SSM meant that `python-dirq` became an optional package, however it is still required by APEL but it was not listed as a dependency presumably as SSM required it, but it actually has its own direct dependency which should be listed.

Add this to the SPEC and any other appropriate locations and double check that there's no other missing dependencies.

tofu-rocketry:
Resolves #211.

- Add python-dirq to lib dependencies. It was removed as a hard
  dependency from SSM so was enforced through that, but it's still
  required by the APEL code.
- Remove apel-ssm from server dependencies as it's only required by the
  client.

### DIFF
--- a/apel.spec
+++ b/apel.spec
@@ -25,7 +25,7 @@ The project is written in Python.
 %package lib
 Summary:        Libraries required for Apel Client, Server and Parsers
 Group:          Development/Languages
-Requires:       MySQL-python python-ldap python-iso8601
+Requires:       MySQL-python, python-ldap, python-iso8601, python-dirq
 Requires(pre):  shadow-utils
 
 %description lib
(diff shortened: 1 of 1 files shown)


gregcorbett:
SSM no required by the server? What's the thinking here?

tofu-rocketry:
> SSM not required by the server? What's the thinking here?

It's not actually used by the server - the client calls SSM functions to send messages, but the server just reads in messages that the receiver daemon has written out.

gregcorbett (approved):
LGTM.

PR #242, opened 2021-05-25
Travis, MySQL and testing fixes

tofu-rocketry:
- Changes to Travis file bring our APEL CI into line with SSM (which was updated more recently).
- MySQL testing has been changed to use MariaDB defaults from 10.1.x as 10.2.x defaults don't work with the current schemas.
- Some fixes to the client schema required by non-ancient versions of MySQL (or just typos).

Lack of update script recorded in https://github.com/apel/apel/issues/243

### DIFF
--- a/schemas/client.sql
+++ b/schemas/client.sql
@@ -140,7 +140,7 @@ CREATE TABLE BlahdRecords (
   VORoleID                    INT          NOT NULL, -- foreign key
   CEID                        INT          NOT NULL, -- foreign key
   GlobalJobId                 VARCHAR(255) DEFAULT NULL,
-  LrmsId                      VARCHAR(255) DEFAULT NULL,
+  LrmsId                      VARCHAR(255) DEFAULT '',
   SiteID                      INT          NOT NULL, -- foreign key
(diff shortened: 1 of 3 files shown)


schemas/client.sql
```diff
@@ -140,13 +140,13 @@ CREATE TABLE BlahdRecords (
   VORoleID                    INT          NOT NULL, -- foreign key
   CEID                        INT          NOT NULL, -- foreign key
   GlobalJobId                 VARCHAR(255) DEFAULT NULL,
-  LrmsId                      VARCHAR(255) DEFAULT NULL,
+  LrmsId                      VARCHAR(255) DEFAULT '',
```
gregcorbett, line 143: Do we want to make an update script for clients? As we've done for [server schema changes](https://github.com/apel/apel/blob/1.9.0-1/scripts/update-1.7.0-1.8.2.sql) in the past?

tofu-rocketry: I've made a note in https://github.com/apel/apel/issues/243

gregcorbett (approved):
LGTM :)

PR #250, opened 2021-06-23
246 no cloudtype primary key

### LINKED ISSUE #246: Remove CloudType from CloudSummaries primary key
From the [GGUS ticket](https://ggus.eu/index.php?mode=ticket_info&ticket_id=151118) (atleast), with CloudType being part of the [CloudSummaries table primary key](https://github.com/apel/apel/blob/dev/schemas/cloud.sql#L181), we end up having stale summaries because the CloudType (i.e. the verison of caso used to extract the records) can change on a republish.

To alleviate this, we could remove CloudType from the primary key. 

Given that CloudComputeServiceID is also in the CloudSummaries primary key and is grouped on by the summariser, I can't think of a down side.

pjcon:
Resolves #246

### DIFF
--- a/scripts/update-x.x.x-x.x.x.sql
+++ b/scripts/update-x.x.x-x.x.x.sql
@@ -22,5 +22,87 @@ ALTER TABLE CloudSummaries
     BenchmarkType, Benchmark
   );
 
+DROP PROCEDURE IF EXISTS SummariseVMs;
+DELIMITER //
+CREATE PROCEDURE SummariseVMs()
+BEGIN
+
+  -- Based on discussion here:
+  -- http://stackoverflow.com/questions/13196190/mysql-subtracting-value-from-previous-row-group-by
+  CREATE TEMPORARY TABLE TVMUsagePerMonth
+    (INDEX index_VMUsagePerMonth USING BTREE (VMUUID, Month, Year))
+    SELECT
+      ThisRecord.RecordCreateTime as RecordCreateTime,
+      ThisRecord.VMUUID as VMUUID,
+      ThisRecord.SiteID as SiteID,
+      ThisRecord.CloudComputeServiceID as CloudComputeServiceID,
+      ThisRecord.MeasurementMonth as Month,
+      ThisRecord.MeasurementYear as Year,
+      ThisRecord.GlobalUserNameID as GlobalUserNameID,
+      ThisRecord.VOID as VOID,
+      ThisRecord.VOGroupID as VOGroupID,
+      ThisRecord.VORoleID as VORoleID,
+      ThisRecord.Status as Status,
+      ThisRecord.CloudType as CloudType,
+      ThisRecord.ImageId as ImageId,
+      ThisRecord.StartTime as StartTime,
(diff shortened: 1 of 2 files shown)


pjcon:
I've not included versions in the schema update script.

scripts/update-x.x.x-x.x.x.sql
```diff
@@ -0,0 +1,61 @@
...
+-- upgrade to APEL Version next, remove the block comment
+-- symbols around this section and run this script
+
+-- This section will:
+-- - Remove CloudType from the CloudSummaries primary key. This
+--   ensures new summaries overwrite old summaries after updates
+--   to CASO version name.
+
+DROP TABLE IF EXISTS CloudSummaries;
+CREATE TABLE CloudSummaries (
```
tofu-rocketry, line 18: We can't drop and recreate an existing table as part of an update schema as that'll drop the data it contains! This should be one or more `ALTER TABLE` statements to change the primary key.

tofu-rocketry (changes requested):

tofu-rocketry:
Note that I've rebased this before you make any additions to this PR.

pjcon (commented):
No longer dropping the table, but instead drop and readd the primary key without CloudType.

tofu-rocketry (approved):
Looks good. Just could do with a note on the manual testing.

gregcorbett (dismissed):
Do we need to remove CloudType from the `SummariseVMs` procedure? If we don't, won't we generate summaries that only differ by CloudType and then only one will get be kept.

tofu-rocketry:
> Do we need to remove CloudType from the `SummariseVMs` procedure? If we don't, won't we generate summaries that only differ by CloudType and then only one will get be kept.

Good point!

tofu-rocketry:
> Do we need to remove CloudType from the `SummariseVMs` procedure? If we don't, won't we generate summaries that only differ by CloudType and then only one will get be kept.

Ah, having a look at Connor's solution, we do want to take it out of the grouping, but we still want to select the CloudType from the later record, so that the latest CloudType is stored in the summaries. So I think we just need to revert the Null bit.

gregcorbett:
> Ah, having a look at Connor's solution, we do want to take it out of the grouping, but we still want to select the CloudType from the later record, so that the latest CloudType is stored in the summaries. So I think we just need to revert the Null bit.

So the idea being: if a site publishes a record with `{Month:8, Year: 2024, CloudType: Foo 1.0.0}` and then later republished with `{Month:8, Year: 2024, CloudType: Foo 1.1.1}` (and all other things being equal) we'd end up with a fresh summary showing `CloudType: Foo 1.1.1` and no stale summary for `CloudType: Foo 1.0.0`.

Is that correct?

If that's the case, I think this looks okay.

tofu-rocketry:
> So the idea being: if a site publishes a record with `{Month:8, Year: 2024, CloudType: Foo}` and then later republished with `{Month:8, Year: 2024, CloudType: Bar}` (and all other things being equal) we'd end up with a fresh summary showing `CloudType: Bar` and no stale summary for `CloudType: Foo`.
> 
> Is that correct?

Yeah, that's the idea. `SummariseVMs` should take care of preserving the later data for us, but we do want to keep the `CloudType` in the Summaries.

gregcorbett (approved):

Issue #259, opened 2021-10-07
Add Collector field to CloudRecords


pjcon:
Currently CloudType has diverged from its original definition which creates extra complexity.

The string currently implies the cloud type the record came from by giving the name/version of the collector software for that cloud e.g., Caso can query OpenStack, so the CloudType field may say "caso 1.4.3" instead of "openstack".

This can be fixed by returning CloudType to its original usage (by requesting the correct string from providers), and creating an extra field to account for the also-useful collector for debugging.

PR #260, opened 2021-10-07
259 add collector field cloudrecords

pjcon:
Fixes issue #259 
Builds on pull-request #250 (issue #246)

I've added the "Collector" field to the CloudRecord type and schema, then returned CloudType to its intended use in CloudSummaries/SummariseVMs

### DIFF
--- a/schemas/cloud.sql
+++ b/schemas/cloud.sql
@@ -78,7 +78,7 @@ CREATE PROCEDURE ReplaceCloudRecord(
   cpuCount INT, networkType VARCHAR(255),  networkInbound INT,
   networkOutbound INT, publicIPCount INT, memory INT,
   disk INT, benchmarkType VARCHAR(50), benchmark DECIMAL(10,3), storageRecordId VARCHAR(255),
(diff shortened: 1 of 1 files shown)


jrha (commented):
I don't understand what `scripts/update-x.x.x-x.x.x.sql` is being added for? It doesn't appear to actually target a specific version?

schemas/cloud.sql
```diff
@@ -284,7 +285,7 @@ BEGIN
       'summariser'
       FROM TVMUsagePerMonth
       GROUP BY SiteID, CloudComputeServiceID, Month, Year, GlobalUserNameID, VOID,
-          VOGroupID, VORoleID, Status, CloudType, ImageId, CpuCount,
```
gregcorbett, line 287: Should we remove `CloudType` from the rest of the `SummariseVMs` procedure now that we aren't grouping by it?

gregcorbett (changes requested):

gregcorbett:
> We should get the FedClouds' thoughts on this as the new field would necessitate a new message version.

I will raise this at the next FedCloud meeting.

PR #263, opened 2021-10-13
Secondary URLs for fetching DNs from multiple GOCDBs

### LINKED ISSUE #77: Use secondary URL when fetching DNs
The auth script, `retrieve_dns`, isn't able to use a secondary GOCDB URL (which is available) if the first fails. Would be handy if it could.

ghost:
Resolves #77

Change the auth configuration section to allow multiple GOCDB hosts to be specified to combine with the GOCDB PI command and service type tags to fetch the list of authorised APEL clients.

Implement code to try each GOCDB host in sequence until a successful connection is made.

### DIFF
--- a/bin/retrieve_dns.py
+++ b/bin/retrieve_dns.py
@@ -44,7 +44,7 @@ class Configuration(object):
     """Dummy class for attaching configuration to."""
     def __init__(self):
         self.gocdb_hosts = None
-        self.gocdb_pi_cmd = None
+        self.gocdb_pi_path = None
         self.extra_dns = None
         self.banned_dns = None
         self.dn_file = None
@@ -65,9 +65,9 @@ def get_config(config_file):
         c.gocdb_hosts = None
 
     try:
-        c.gocdb_pi_cmd = cp.get('auth', 'gocdb_pi_cmd')
+        c.gocdb_pi_path = cp.get('auth', 'gocdb_pi_path')
     except ConfigParser.NoOptionError:
-        c.gocdb_pi_cmd = None
+        c.gocdb_pi_path = None
 
     try:
         c.service_types = cp.get('auth', 'service_types')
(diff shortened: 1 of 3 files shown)


jrha:
I've rebased this on `dev` and dealt with the conflicts, which were just related to logging statements.

jrha:
This appears to be based on #214, so that should probably be reviewed and merged first.

bin/retrieve_dns.py
```diff
@@ -58,9 +60,21 @@ def get_config(config_file):
...
+        c.gocdb_pi_cmd = cp.get('auth', 'gocdb_pi_cmd')
+    except ConfigParser.NoOptionError:
+        c.gocdb_pi_cmd = None
+
+    try:
+        c.service_types = cp.get('auth', 'service_types')
+    except ConfigParser.NoOptionError:
+        c.service_types = None
+    if c.service_types and not verify_service_types(c.service_types):
+        raise ValueError("Invalid characters in service type.")
```
gregcorbett, line 77: If `service_types` was missing, or present but empty, in the config - would that cause `c.service_types` to be `None` and a `ValueError` to be raised?

If so, the error message should be change to better explain those cases.

gregcorbett (changes requested):
Generally looks good, I've left two comments.

bin/retrieve_dns.py
gregcorbett: ah yes, I must have been confused...

bin/retrieve_dns.py
```diff
@@ -42,7 +43,8 @@
 class Configuration(object):
     """Dummy class for attaching configuration to."""
     def __init__(self):
-        self.gocdb_url = None
+        self.gocdb_hosts = None
+        self.gocdb_pi_cmd = None
```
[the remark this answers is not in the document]
ineilson: Rename name throughout and pushed changes.

gregcorbett (approved):

gregcorbett:
I feel like the "Inefficient regular expression" issues CodeQL has identified is one for GOCDB, see https://github.com/GOCDB/gocdb/issues/379.

tofu-rocketry:
Reopened as #302

PR #264, opened 2021-12-02
Add gpu records

pjcon:
This PR implements GPURecords and GPUSummaries to account for GPU usage in APEL.

Changes in this pull request:

- [x] Implements GPURecord/Summary classes
- [x] JSON GPURecord/Summary schemas
- [x] SQL GPURecord/Summary schemas 
- [x] Loader accepts JSON message format
- [ ] ~~Loader accepts APEL message format~~
- [x] AcceleratorRecord test coverage
- [x] AcceleratorSummary test coverage

### DIFF
--- a/apel/common/json_utils.py
+++ b/apel/common/json_utils.py
@@ -1,0 +1,35 @@
+import json
+
+JSON_MSG_BOILERPLATE = """{
+    "Type": "%s",
+    "Version": "%s",
+    "UsageRecords": %s
+}"""
+
+
+def to_string(json_dict, dlm=''):
+    """ Convert a dict into a valid json string """
+    json_string = str(json_dict).replace("'", '"')
+
+    try:
+        json_dict_conv = json.loads(json_string)
+    except json.decoder.JSONDecodeError:
+        raise Exception('Unable to generate valid JSON. Make sure speech marks or apostrophes are escaped correctly.')
+
+    if json_dict != json_dict_conv:
+        raise Exception('JSON conversion unsuccessful because input dict does not match reconstructed dict.')
+
+    return json_string
+
+def to_dict(json_string):
+    """ Convert a valid json string into a dict """
+
+    try:
+        json_dict = json.loads(json_string)
+    except json.decoder.JSONDecodeError as e:
+        print(e)
+        raise Exception(e, '\n\nJSON string is not valid.')
+
+def to_message(type, version, *usage_records):
+    """ Generate JSON message format """
+    return (JSON_MSG_BOILERPLATE % (type, version, list(usage_records))).replace("'", '"')
--- a/apel/db/loader/loader.py
+++ b/apel/db/loader/loader.py
@@ -199,6 +199,7 @@ class Loader(object):
                     apel.db.records.cloud_summary.CloudSummaryRecord:
                     'Cloud Summary',
                     apel.db.records.accelerator.AcceleratorRecord: 'Accelerator',
+                    apel.db.records.accelerator_summary.AcceleratorSummary: 'Accelerator Summary',
                     }
 
         log.info('Loading message from %s', signer)
--- a/apel/db/loader/loader.py
+++ b/apel/db/loader/loader.py
@@ -197,7 +197,9 @@ class Loader(object):
                     apel.db.records.sync.SyncRecord: 'Sync',
                     apel.db.records.cloud.CloudRecord: 'Cloud',
                     apel.db.records.cloud_summary.CloudSummaryRecord:
-                    'Cloud Summary'}
+                    'Cloud Summary',
+                    apel.db.records.accelerator.AcceleratorRecord: 'Accelerator',
+                    }
 
         log.info('Loading message from %s', signer)
 
--- a/apel/db/loader/record_factory.py
+++ b/apel/db/loader/record_factory.py
@@ -18,15 +18,18 @@
 Module containing the RecordFactory class.
 '''
 
+from apel.common.message_schemas import ACCELERATOR_MSG_SCHEMA
 from apel.db.records.job import JobRecord
 from apel.db.records.summary import SummaryRecord
 from apel.db.records.normalised_summary import NormalisedSummaryRecord
 from apel.db.records.sync import SyncRecord
 from apel.db.records.cloud import CloudRecord
 from apel.db.records.cloud_summary import CloudSummaryRecord
+from apel.db.records.accelerator import AcceleratorRecord
 from apel.db import (JOB_MSG_HEADER, SUMMARY_MSG_HEADER,
                      NORMALISED_SUMMARY_MSG_HEADER, SYNC_MSG_HEADER,
-                     CLOUD_MSG_HEADER, CLOUD_SUMMARY_MSG_HEADER)
+                     CLOUD_MSG_HEADER, CLOUD_SUMMARY_MSG_HEADER,
+                     ACCELERATOR_MSG_TYPE)
 
 from apel.db.loader.car_parser import CarParser
 from apel.db.loader.aur_parser import AurParser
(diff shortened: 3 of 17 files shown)


gregcorbett:
It occurs to me when we designed the record, we tried to do it in a generic way that could support different types of "accelerators" - i.e. GPUs, FPGAs, etc.

However when it came to mocking up example messages they all had `type: APEL GPU message` (as did the corresponding JSON schema). This probably should have been `type: APEL Accelerator message` , maybe we should refer to them as accelerators in the code/database?

pjcon:
I can change all references from GPU to Accelerator where you've mentioned, and also "GPURecord" -> "AcceleratorRecord".

tofu-rocketry:
> It occurs to me when we designed the record, we tried to do it in a generic way that could support different types of "accelerators" - i.e. GPUs, FPGAs, etc.
> 
> However when it came to mocking up example messages they all had `type: APEL GPU message` (as did the corresponding JSON schema). This probably should have been `type: APEL Accelerator message` , maybe we should refer to them as accelerators in the code/database?

What actually got agree with FedCloud though?

gregcorbett:
> What actually got agreed with FedCloud though?

That the record should be generic enough to support non GPU accelerators.

schemas/accelerator.sql
```diff
@@ -0,0 +1,209 @@
...
+  UpdateTime TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  MeasurementMonth INT NOT NULL,
+  MeasurementYear INT NOT NULL,
+
+  AssociatedRecordType VARCHAR(255) NOT NULL,
+  AssociatedRecord VARCHAR(255) NOT NULL,
+
+  GlobalUserName VARCHAR(255),      
+  FQAN VARCHAR(255) NOT NULL,       
+  SiteName VARCHAR(255) NOT NULL,
```
gregcorbett, line 15: `SiteName VARCHAR(255) NOT NULL` should be `SiteID INT NOT NULL` to be consistent with other forms of APEL accounting (both in the `AcceleratorRecords` and `AcceleratorSummaries` tables).

@tofu-rocketry, was the intention the accelerators would be an "add-on" to cloud or grid accounting, or its own standalone type of accounting? If it needs to be standalone - a `Sites` table and `SitesLookup` function would need to be added similar to https://github.com/apel/apel/blob/1.9.2-1/schemas/server.sql#L420

apel/common/json_utils.py
```diff
@@ -0,0 +1,35 @@
+import json
+
+JSON_MSG_BOILERPLATE = """{
+    "Type": "%s",
+    "Version": "%s",
+    "UsageRecords": %s
+}"""
+
+
+def to_string(json_dict, dlm=''):
```
Sae126V, line 10: Is this for future use ?

apel/db/loader/loader.py
```diff
@@ -197,7 +197,10 @@ def load_msg(self, msg_text, signer):
                     apel.db.records.sync.SyncRecord: 'Sync',
                     apel.db.records.cloud.CloudRecord: 'Cloud',
                     apel.db.records.cloud_summary.CloudSummaryRecord:
-                    'Cloud Summary'}
+                    'Cloud Summary',
+                    apel.db.records.accelerator.AcceleratorRecord: 'Accelerator',
+                    apel.db.records.accelerator_summary.AcceleratorSummary: 'Accelerator Summary',
```
Sae126V, line 202: Do we need trailing comma ?

apel/db/loader/record_factory.py
```diff
@@ -115,6 +149,43 @@ def create_records(self, msg_text):
     # Private methods below
     ######################################################################
 
+    def _validate_json(self, js, schema):
+        '''Run jsonschema validate on JSON given a schema otherwise raise a friendly error'''
```
Sae126V, line 153: Should we need to convert single quotes to double quotes across code

schemas/accelerator.sql
```diff
@@ -0,0 +1,209 @@
...
+    REPLACE INTO AcceleratorSummaries(Month, Year, AssociatedRecordType,
+        GlobalUserName, SiteName, 
+        Cores, Count, AvailableDuration, ActiveDuration, 
+        BenchmarkType, Benchmark, Type, Model, NumberOfRecords, PublisherDN)
+    SELECT 
+      MeasurementMonth, MeasurementYear,
+      AssociatedRecordType,
+      GlobalUserName,
+      SiteName,
+      Count,
```
Sae126V, line 131: Should this to Cores followed by Count ?

schemas/accelerator.sql
```diff
@@ -0,0 +1,209 @@
...
+  siteName,
+  count,
+  cores,
+  activeDuration,
+  availableDuration,
+  benchmarkType,
+  benchmark,
+  type,
+  model,
+  DNLookup(publisherDN)
```
Sae126V, line 89: Is `DNLookup` Function missing here ?

schemas/accelerator.sql
```diff
@@ -0,0 +1,209 @@
...
+      Count,
+      Cores,
+      SUM(AvailableDuration),
+      SUM(ActiveDuration),
+      BenchmarkType,
+      Benchmark,
+      Type,
+      Model,
+      COUNT(*),
+      'summariser'
```
Sae126V, line 140: If we are using this 'summariser' to every row. As this being repeated shall we pass this data when unloading ? or Do we need it in the database ?

schemas/accelerator.sql
```diff
@@ -0,0 +1,209 @@
...
+
+DROP PROCEDURE IF EXISTS SummariseAccelerators;
+DELIMITER //
+CREATE PROCEDURE SummariseAccelerators()
+
+BEGIN
+    REPLACE INTO AcceleratorSummaries(Month, Year, AssociatedRecordType,
+        GlobalUserName, SiteName, 
+        Cores, Count, AvailableDuration, ActiveDuration, 
+        BenchmarkType, Benchmark, Type, Model, NumberOfRecords, PublisherDN)
```
Sae126V, line 125: As we are sending 'summariser' to PublisherDN. Should the name be changed ?

schemas/accelerator.sql
```diff
@@ -0,0 +1,209 @@
...
+  siteName VARCHAR(255),
+  count DECIMAL(10,3),
+  cores INT,
+  activeDuration INT,
+  availableDuration INT,
+  benchmarkType VARCHAR(255),
+  benchmark DECIMAL,
+  type VARCHAR(255),
+  model VARCHAR(255),
+  publisherDN VARCHAR(255)
```
Sae126V, line 52: publisherDN INT or publisherDN VARCHAR(255) ?

schemas/accelerator.sql
```diff
@@ -0,0 +1,209 @@
...
+  SiteName,
+  Count,
+  Cores,
+  ActiveDuration,
+  AvailableDuration,
+  BenchmarkType,
+  Benchmark,
+  Type,
+  Model,
+  PublisherDNID
```
Sae126V, line 71: Should this be PublisherDN ?

apel/db/loader/loader.py
tofu-rocketry: Yes, I generally prefer trailing commas in multi-line data structures as it means future additions don't change the last line.

apel/db/loader/record_factory.py
tofu-rocketry: Double quotes would be better.

apel/common/json_utils.py
tofu-rocketry: Removed in #348

tofu-rocketry:
Closed in favour of #459

PR #273, opened 2022-02-16
Ceph StAR storage accounting record generation script

jounaidr:


jounaidr:
Mark as draft for now as changes to handle `measureTime` and `validDuration` are going to be done inside apel rather than in the script.

gregcorbett (changes requested):
Generally I think this looks good. It would be interesting to see what codeclimate has to say about. I tired to trigger a re-analysis but it doesn't seem to have worked.

I also tried rebasing the PR locally (thinking a rebase and push would also trigger codeclimate) but it was non-trivial and I was actively making the commit history worse.

@tofu-rocketry, we could try the "Update with merge commit" option in the GitHub UI?

PR #285, opened 2022-06-21
Gap unload for summary records

### LINKED ISSUE #225: Add ability to `gap` unload for summaries
Usually unload `all` is ok for sites to do as it's not a huge amount of data, but for larger sites it can be. It is also a "gotcha" for users who expect that each method of unloading should be useable for each type of record and don't always notice that their republish of summaries didn't actually work.

So, it would be useful to add an unload `gap` option for summaries.

ineilson:
Fixes #225 

I'm not really convinced of the usefulness of the test_gap_unload_summaries test, without using a dummy dataset, but maybe better than no test at all.


apel/db/unloader.py
```diff
@@ -119,8 +119,8 @@ def unload_gap(self, table_name, start, end, ur=False):
         '''
         record_type = self.RECORD_TYPES[table_name]
 
-        if record_type != JobRecord:
-            raise ApelDbException("Can only gap publish for JobRecords.")
+        if record_type != JobRecord and record_type != SummaryRecord:
```
tofu-rocketry, line 122:
```suggestion
        if record_type not in (JobRecord, SummaryRecord):
```
This is the tidier way to do this.

test/test_unloader.py
```diff
@@ -3,6 +3,8 @@
 import tempfile
 import unittest
 
+from mock import mock
```
tofu-rocketry, line 6:
```suggestion
from mock import mock

```
Could do with a line break to divide the 3rd-party modules from our own.

test/test_unloader.py
```diff
@@ -3,6 +3,8 @@
...
+from apel.db.backends.mysql import ApelMysqlDb
```
tofu-rocketry, line 7: Alphabetize imports please. `backends` after `apledb`.

test/test_unloader.py
```diff
@@ -16,7 +18,15 @@ class TestDbUnloader(unittest.TestCase):
     def setUp(self):
         # Create temporary directory for dirq to write to
         self.dir_path = tempfile.mkdtemp()
-        self.unloader = DbUnloader(None, self.dir_path)
+
+        # Mock out the database as we're not testing that here
+        # Note that as it's a class it needs instantiating
+        self.mock_db = mock.patch(__name__ + '.ApelMysqlDb', autospec=True, spec_set=True)
```
tofu-rocketry, line 24: What's the purpose of `__name___` here?

tofu-rocketry (changes requested):

test/test_unloader.py
ineilson: Without it it fails `TypeError: Need a valid target to patch. You supplied: 'ApelMysqlDb'`. I think this is consistent with https://docs.python.org/3/library/unittest.mock.html#where-to-patch (ok, that's python3), but I'm not so very confident.

PR #294, opened 2022-08-01
Delete accepted messages by default

### LINKED ISSUE #261: Change default of save_messages to false
As it's mainly used for debugging, this should be set to false by default.

https://github.com/apel/apel/blob/1.9.0-1/conf/loader.cfg#L12

ineilson:
Fixes #261


tofu-rocketry (approved):
Perfect!

PR #299, opened 2022-11-28
Add check for presence of config file

rowan04:
- gives a clearer error message if the config file does not exist
- added default paths to the help output on the command line for easier use
- added default path variables for easier future editing and reading


bin/client.py
```diff
@@ -278,25 +278,42 @@ def main():
     install_exc_handler(default_handler)
     ver = 'APEL client %s.%s.%s' % __version__
     opt_parser = OptionParser(version=ver, description=__doc__)
+    default_conf_location = '/etc/apel/client.cfg'
```
tofu-rocketry, line 281: As these are constants, they should be in ALL_CAPS.
https://peps.python.org/pep-0008/#constants

bin/client.py
```diff
@@ -278,25 +278,42 @@ def main():
...
 
-    scp = ConfigParser.ConfigParser()
-    scp.read(options.ssm_config)
+    # check if config file exists using os.path.isfile fuction
+    if os.path.isfile(options.config):
+        ccp = ConfigParser.ConfigParser()
+        ccp.read(options.config)
+    else:
+        print("Config file not found at", options.config)
+        exit(1)
```
tofu-rocketry, line 308: This should be a `sys.exit`.

https://stackoverflow.com/questions/6501121/difference-between-exit-and-sys-exit-in-python

tofu-rocketry (changes requested):

bin/client.py
rowan04: done

bin/client.py
rowan04: done

rowan04:
updated branch to latest dev

tofu-rocketry:
> updated branch to latest dev

Can you use "Update with rebase", rather than "with merge commit" please?

bin/client.py
```diff
@@ -278,25 +278,42 @@ def main():
...
+                          default=DEFAULT_LOGCONF_LOCATION)
 
     options, unused_args = opt_parser.parse_args()
-    ccp = ConfigParser.ConfigParser()
-    ccp.read(options.config)
 
-    scp = ConfigParser.ConfigParser()
-    scp.read(options.ssm_config)
+    # check if config file exists using os.path.isfile fuction
+    if os.path.isfile(options.config):
```
tofu-rocketry, line 303: Could you refactor this into a function and then just call it with `config` and `ssm_config` (to replace the check at line here: https://github.com/apel/apel/pull/299/files#diff-3a1e2d7257907d68bb0bd07963020f8d03268ea7a0d680abebf80278327718d4R311 )?

tofu-rocketry (changes requested):

bin/client.py
rowan04: done

rowan04:
Updated with rebase

rowan04:
Fix typo in commit name

PR #300, opened 2022-12-12
Add sanity checks on start and end times for cloud records at load time.

### LINKED ISSUE #239: Add sanity checks on start and end times for cloud records at load time.
There are no sanity checks on the value of StartTime and EndTime when we load a cloud record from a message. 

Some sensible sanity checks may include
- The record has a StartTime
- If the record has an EndTime / is in the "completed" state, then the EndTime comes after the StartTime
- If the record has an EndTime / is in the "completed" state, then the EndTime is not in the future.

It looks like [some code](https://github.com/apel/apel/blob/1.9.0-1/apel/db/records/cloud.py#L107) exists as an attempt to do this, but the check isn't called as of version 1.9.0. This code may have been disabled to allow development/older versions of accounting collectors to publish cloud records. We have certainly received and loaded messages from pre 1.4.X versions of caso that omit both the start and end times.

Perhaps a starting point is to add StartTime as a mandatory field in a new version of the cloud record, as we can't think of a reason why this should be omitted.

gregcorbett:
Resolves #239

### DIFF
--- a/apel/db/records/cloud.py
+++ b/apel/db/records/cloud.py
@@ -100,19 +100,18 @@ class CloudRecord(Record):
         if self._record_content['CpuCount'] is None:
             self._record_content['CpuCount'] = 0
 
-        # Sanity check the values of StartTime and EndTime
+        # Sanity check the values of StartTime and EndTime.
         self._check_start_end_times()
+        # Sanity check the values of Status and EndTime.
+        self._check_endtime_status_combination()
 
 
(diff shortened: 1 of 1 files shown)


gregcorbett:
The first 4 commits are only vaguley related to this PR, as when started the issue I was running Python3, and then I tried to improve the files I had touched as part of the PR.

Happy to factor out into a seperate PR if prefered.

PR #302, opened 2023-05-05
Ian's secondary urls

### LINKED ISSUE #77: Use secondary URL when fetching DNs

tofu-rocketry:
Replaces closed PR #263 as fork has been deleted.

Resolves #77

Change the auth configuration section to allow multiple GOCDB hosts to be specified to combine with the GOCDB PI command and service type tags to fetch the list of authorised APEL clients.

Implement code to try each GOCDB host in sequence until a successful connection is made.

### DIFF
--- a/bin/retrieve_dns.py
+++ b/bin/retrieve_dns.py
@@ -260,6 +260,12 @@ def verify_dn(dn):
     return True
 
 def verify_service_types(service_types):
+    '''
+    Returns true if all the provided service types are valid GOCDB service types.
+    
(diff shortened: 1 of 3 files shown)


gregcorbett (approved):
Looks the same as **https://github.com/apel/apel/pull/263** to me, which itself looked good to me.

PR #306, opened 2023-05-24
Update default hostname for BDII

tofu-rocketry:
- CERN BDII is being decommissioned so this changes it to the EGI one which is still available.

As requested by CERN ops.


rowan04 (approved):
lgtm

gregcorbett (approved):
LGTM

PR #307, opened 2023-05-24
Add update schema

### LINKED ISSUE #243: Create update schema for > 1.8.2
An update schema wasn't created for 1.8.2 to 1.9.0 for this https://github.com/apel/apel/pull/227

There's also a change to a default in the client schema which will need an update schema: https://github.com/apel/apel/pull/242

tofu-rocketry:
Resolves #243 

Manually tested by comparing output of `SHOW CREATE TABLE` from an upgraded schema to the 1.9.1 schema.

### DIFF
--- a/scripts/update-1.8.2-1.9.1.sql
+++ b/scripts/update-1.8.2-1.9.1.sql
@@ -1,0 +1,14 @@
+-- This script contains a SQL block that can update
+-- APEL version 1.8.2 databases of the following types to 1.9.1:
+--  - Client Grid Accounting Database
+
+-- UPDATE SCRIPT FOR CLIENT SCHEMA
+
+-- This will set any null LrmsIds to the empty string and change the default value to the same.
+
+UPDATE BlahdRecords SET LrmsId = '' WHERE LrmsId is NULL;
+ALTER TABLE BlahdRecords MODIFY COLUMN LrmsId VARCHAR(255) DEFAULT '';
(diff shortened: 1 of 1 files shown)


rowan04 (approved):
lgtm

gregcorbett:
Should the update script include this change https://github.com/apel/apel/commit/3100a5acac57eced6c50e0cc3065581a29af6149 ?

gregcorbett:
Should the script reference version 1.9.2 rather than 1.9.1?

tofu-rocketry:
> Should the update script include this change [3100a5a](https://github.com/apel/apel/commit/3100a5acac57eced6c50e0cc3065581a29af6149) ?

The spelling mistake was in the `DROP` statement, so there isn't actually anything to add to an update schema whose purpose is to bring an existing schema up to date as the `DROP` doesn't "persist" in an applied schema.

(If you re-applied the schema, the `DROP` would do nothing and the `CREATE` would potentially fail.)

tofu-rocketry:
> ~Should the script reference version 1.9.2 rather than 1.9.1?~ Ah nevermind, get the context of this now. This script should have been added in version 1.9.1, but wasn't - so is being added now.

Yeah, I did go round the houses with that one, but came down on the side of it referring to the version that has the schema change.

gregcorbett (approved):
LGTM

PR #315, opened 2023-06-19
[GT-159] Add YAML support for summarised SR's

Sae126V:
Build on top of https://github.com/apel/apel/pull/229. 

This [[GT-159]] helps us to dump the summarised storage records into YAML format.

"Resolves #229 "

[GT-159]: https://stfc.atlassian.net/browse/GT-159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apel/db/unloader.py
```diff
@@ -207,10 +211,14 @@ def _write_messages(self, record_type, table_name, query, ur):
...
-                self._write_xml(batch)
+
+            if table_name == 'DaySummaries':
+                self._write_yaml(batch)
             else:
-                self._write_apel(batch)
+                if ur:
+                    self._write_xml(batch)
+                else:
+                    self._write_apel(batch)
```
gregcorbett, line 221: I wonder if this would be better as

```
if table_name == 'DaySummaries':
   ...
elif ur:
   ...
else:
    ...
```

Sae126V: Well. Actually, I didn't want to disturb or change the existing flow of the code. The changes I added will not impact the pervious code :). Hope that make sense ?

apel/db/unloader.py
```diff
@@ -271,6 +279,33 @@ def _write_apel(self, records):
...
+        documents = []
+
+        if type(records[0]) == StorageSummaryRecord:
+            documents.append({'type': 'StorageRecord'})
+
+            for record in records:
+                documents.append(record.get_ur(self._withhold_dns))
+        else:
+            raise ApelDbException(
+                "Can only send record's in YAML for StorageRecords"
```
rowan04, line 297:
```suggestion
                "Can only send records in YAML for StorageRecords"
```
grammar error i think

apel/db/backends/mysql.py
```diff
@@ -293,6 +294,29 @@ def check_duplicate_sites(self):
...
+        '''
+        try:
+            # prevent MySQLdb from raising
+            # 'MySQL server has gone' exception
+            self._mysql_reconnect()
+
+            c = self.db.cursor()
+
+            log.info("Summarising storage records...")
+            c.callproc(self._summarise_storage_proc, ())
```
tofu-rocketry, line 309: I realise it's in the other similar methods, but what's the point of the empty brackets as an argument?

apel/db/records/storage_summary.py
```diff
@@ -0,0 +1,126 @@
...
+
+import logging
+
+# get the relevant logger
+log = logging.getLogger(__name__)
+
+
+class StorageSummaryRecord(Record):
+    '''
+    Class to represent one `summarised` storage record.
```
tofu-rocketry, line 28: I don't think we need the back-ticks around "summarised", do we?

schemas/storage.sql
```diff
@@ -374,3 +374,49 @@ WHERE StarRecords.StorageSystemID = StorageSystems.id
...
+-- -----------------------------------------------------------------------------
+-- DaySummaries
+DROP TABLE IF EXISTS DaySummaries;
+CREATE TABLE DaySummaries (
+  Site                        VARCHAR(255),
+  StorageSystem               VARCHAR(255),
+  StorageShare                VARCHAR(255),
+  StorageMedia                VARCHAR(255),
+  DirectoryPath               VARCHAR(255),
+  `Group`                     VARCHAR(255),
```
rowan04, line 387: Why the backticks here? Codacy doesn't seem to like it

apel/db/records/storage_summary.py
```diff
@@ -0,0 +1,126 @@
...
+        The source (DN of the sender) isn't used in this record type currently.
+        """
+        return Record.get_db_tuple(self)
+
+    def verify_and_add_field(self, doc, field_key, value_converter=None):
+        '''
+        Helper method to verify if a field is NOT None.
+
+        The `field_key` parameter is the name of the field in the record.
+        The `value_converter` parameter is an optional for specifiying type,
```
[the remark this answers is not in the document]
Sae126V: Done.. 👍

apel/db/unloader.py
Sae126V: Right. Updated the code now.

apel/db/backends/mysql.py
```diff
@@ -86,6 +86,7 @@ def __init__(self, host, port, username, pwd, db):
         self._db_name = db
 
         self._summarise_jobs_proc = "SummariseJobs"
+        self._summarise_storage_proc = "DailySummary"
```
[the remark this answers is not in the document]
Sae126V: Changed the DailySummary to SummariseStorageRecords. Done

apel/db/backends/mysql.py
Sae126V: You are right. In our case, Passing an empty args() to proc is not needed now. Calling the proc directly now. Done

apel/db/records/storage_summary.py
```diff
@@ -0,0 +1,126 @@
...
+from apel.db.records import Record
+
+import logging
+
+# get the relevant logger
+log = logging.getLogger(__name__)
+
+
+class StorageSummaryRecord(Record):
+    '''
```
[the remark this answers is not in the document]
Sae126V: Sure @tofu-rocketry . I wasn't sure which one to use before. I am clear now. The document you have shared also helped me .

apel/db/records/storage_summary.py
Sae126V: I have used it to differentiate and stress more on `summarised` so that one can understand what it is referring to like Summarised records. It make sense that summarised alone should help us understanding it's an summarised storage records. So, I have removed the backticks.

schemas/storage.sql
Sae126V: We might need to give more priority to SQL here than Codacy. @rowan04, As we wanted column name to be Group. We are using as `Group` because since it is a reserve keyword. We cannot directly use keywords because it might throw syntax errors or confuses SQL.

apel/db/records/storage_summary.py
```diff
@@ -0,0 +1,126 @@
...
+        We have to go back to the apel_db object to find the stored procedure.
+        This is because only this object knows what type of record it is,
+        and only the apel_db knows what the procedure details are.
+        """
+
+        values = self.get_db_tuple(source)
+
+        return values
+
+    def get_db_tuple(self, source=None):
```
rowan04, line 74: why need source in the parameters if it isn't being used?

test/test_storage_summary.py
```diff
@@ -0,0 +1,80 @@
...
+    def test_load_from_tuple(self):
+        # full example
+        data = ('MySite', 'host.example.org', 'pool-003', 'disk',
+                '/home/projectA', 'binarydataproject.example.org',
+                '2023-10-11T09:41:40Z', 10, 2023, 42, 14728, 13617, 0)
+
+        record = StorageSummaryRecord()
+        record.load_from_tuple(data)
+
+        self.assertEquals(record.get_field('Site'), 'MySite')
```
rowan04, line 20: As of Python 2.7, assertEquals() is being deprecated in favour of assertEqual() - https://docs.python.org/2/library/unittest.html#deprecated-aliases

apel/db/records/storage_summary.py
Sae126V: You are right. As of now, we aren't making use of this feature to determine which DN. I have referred to other records while writing this storage record. I added the 'source' argument to maintain consistency with other records and to inform others about its existence for potential use in the future. This way, they will know that there is a feature called 'Source' available, should it be needed.

test/test_storage_summary.py
Sae126V: Cool, Updated the code. I think it would be good chance for us to have a ticket for this please ? what's say ?

rowan04: Could do. Wait for Adrian/Greg to confirm they would want it though

PR #316, opened 2023-08-07
Update to use except..as from commas and fix print

tofu-rocketry:
A few Python 3 type fixes to support HEPscore changes


rowan04 (approved):
lgtm

PR #322, opened 2023-09-18
Add accelerator records and summaries

Sae126V:
Minor changes to #264 . Once summarizes the accelerator records and user performs unloading.
The output would be:
```

APEL-Accelerator-summary-message
Month: 9
Year: 2023
AssociatedRecordType: cloud
GlobalUserName: RALUserA
SiteName: RAL-SITE
Count: 1.4
Cores: 128
AvailableDuration: 4000
ActiveDuration: None
BenchmarkType: HEPSPEC
Benchmark: 9000.0
Type: GPU
Model: ModelA
NumberOfRecords: 1
PublisherDNID: summariser
%%
```

### DIFF
--- a/schemas/accelerator.sql
+++ b/schemas/accelerator.sql
@@ -91,3 +91,119 @@ VALUES(
 END //
 DELIMITER ;
 
+
+DROP TABLE IF EXISTS AcceleratorSummaries;
+CREATE TABLE AcceleratorSummaries (
+    Month INT NOT NULL, 
+    Year INT NOT NULL,
+    AssociatedRecordType VARCHAR(255) NOT NULL,
+    GlobalUserName VARCHAR(255), 
+    SiteName VARCHAR(255) NOT NULL, 
+    Count DECIMAL(10,3) NOT NULL,
+    Cores INT,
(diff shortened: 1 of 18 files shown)


schemas/accelerator.sql
```diff
@@ -0,0 +1,218 @@
...
+  AssociatedRecordType VARCHAR(255) NOT NULL,
+  AssociatedRecord VARCHAR(255) NOT NULL,
+
+  GlobalUserName VARCHAR(255),      
+  FQAN VARCHAR(255) NOT NULL,       
+  SiteID INT NOT NULL,
+  Count DECIMAL(10,3) NOT NULL,
+  Cores INT,
+  ActiveDuration INT,
+  AvailableDuration INT,
```
RedProkofiev, line 30: This should not be nullable if this is a mandatory field - also, within the acceleratorrecords schema I see inconsistencies to the mandatory fields in the acceleratorrecord messaging, could we discuss?

Sae126V: I have raised this PR to address a few issues with the code logic, etc. I have not worked with the team to determine which fields are mandatory or not. Sure, we can have a discussion.

PR #323, opened 2023-09-18
[#238] Add storage record type for logger

### LINKED ISSUE #238: Logging of type of record loaded doesn't work for Storage
Following #209, the logging of the type of record loaded does seem to work for Cloud and maybe Grid records (no grid records yet), but not for Storage records - this may be due to the slightly weirder setup that storage accounting has with two record types.

Sae126V:
When user sends storage record starting with '<'. Previously it doesn't have the class type in `record_type`. Now it should say Storage if it is a storage record when it receives a file from receiver.

Resolves #238

### DIFF
--- a/test/test_loader.py
+++ b/test/test_loader.py
@@ -143,8 +143,9 @@ class LoaderTest(unittest.TestCase):
             mock_log.assert_has_calls(
                         [call('Message contains 1 %s record', 'Summary')])
 
-    def test_load_all_other_type(self):
-        """Check that load_records is called and message is logged correctly when type is other"""
+    def test_load_all_storage_type(self):
+        """Check that load_records is called and message is logged
+            correctly when record type is storage"""
         logger = logging.getLogger('loader')
         pidfile = os.path.join(self.dir_path, 'pidfile')
 
--- a/test/test_loader.py
+++ b/test/test_loader.py
@@ -178,7 +178,7 @@ class LoaderTest(unittest.TestCase):
             self.loader.load_all_msgs()
             self.mock_db.load_records.assert_called_once()
             mock_log.assert_has_calls(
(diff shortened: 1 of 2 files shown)


test/test_loader.py
```diff
@@ -178,7 +178,7 @@ def test_load_all_other_type(self):
             self.loader.load_all_msgs()
             self.mock_db.load_records.assert_called_once()
             mock_log.assert_has_calls(
-                        [call('Message contains %i records', 1)])
+                        [call('Message contains 1 %s record', 'Storage')])
```
rowan04, line 181:
```suggestion
                        [call('Message contains 1 Storage record')])
```
why do we need the %s? Why not just this?

Sae126V: Ah right. So, We are just mocking and expecting that it would be called something like this `[call('Message contains 1 %s record', 'Storage')])` and we are not checking how it would be evaluated @rowan04 . As you suggested I agree that end result we would be getting is as you suggested. Since we are not checking the actual log. We just expect it would be calling something like this ` [call('Message contains 1 %s record', 'Storage')])`. Does that make sense ?

rowan04: yeah my bad, i see now

gregcorbett:
Maybe we should rename the test method `test_load_all_other_type` to `test_load_all_storage_type` (or similar) and update the docstring? 

`test_load_all_other_type` is now no longer checking that a message is logged correctly when type is other - as the type is now `Storage`.

Sae126V:
> `test_load_all_other_type`

Yep Make sense, Greg. Nice catch. I was only worried to run the test cases before. Fixed it now :)

tofu-rocketry (approved):

PR #326, opened 2023-09-26
Reduce parser log verbosity

### LINKED ISSUE #324: Excessive parser logs from reporting unparsed files
The `parser` currently reports on *all* files in the source folder at the standard log level. Critically, that includes files which do not match the pattern - say `blah` files when looking for `batch` files - as well as previously handled files.

https://github.com/apel/apel/blob/0f888108a90298ab2bdc3ed0a3d0c32575ab84ea/bin/parser.py#L215-L221

In practice, for a setup where `blah` and `batch` files are stored together *every file* is reported *twice* on *every run* of the parser. Even for a short service lifetime (~week) that is already dozens to hundreds of useless log lines; for long service lifetime, we have multiple 10k such lines and even short log *windows* (~week) are hundreds of MB in size.

Since the input files are usually created by automated processes, IMO there is little point in reporting them again and again once accounting has been tested. It only makes the log files much less useful to look through and limits the timespan one can store of these logs.

-------------------

Unless a more intricate strategy is desired, I recommend to simply log these messages at `DEBUG` level.

maxfischer2781:
This PR reduces the verbosity of `parser` logs by logging already- and never-matched file paths at DEBUG instead of INFO level.

Closes #324.

### DIFF
--- a/bin/parser.py
+++ b/bin/parser.py
@@ -213,12 +213,12 @@ def scan_dir(parser, dirpath, reparse, expr, apel_db, processed):
                         pr.set_field('Parsed', parsed)
                         updated.append(pr)
                 elif unparsed:
-                    log.info('Skipping file (failed to parse previously): %s',
+                    log.debug('Skipping file (failed to parse previously): %s',
                              abs_file)
                 else:
-                    log.info('Skipping file (already parsed): %s ', abs_file)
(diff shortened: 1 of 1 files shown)


tofu-rocketry:
We did some tweaks and merged it as part of #334.

PR #329, opened 2023-09-26
Update pre-commit config

tofu-rocketry:
- Add a couple of extra tests.
- Update pygrep hook to latest.
- Add config for CI.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,6 +15,7 @@ repos:
   - id: check-added-large-files
   - id: check-merge-conflict
   - id: check-yaml
+  - id: debug-statements
   - id: end-of-file-fixer
   - id: mixed-line-ending
     name: Force line endings to LF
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -15,15 +15,26 @@ repos:
...
+  - id: debug-statements
   - id: end-of-file-fixer
   - id: mixed-line-ending
     name: Force line endings to LF
     args: ['--fix=lf']
   - id: trailing-whitespace
 
 - repo: https://github.com/pre-commit/pygrep-hooks
-  rev: v1.9.0
+  rev: v1.10.0
```
gregcorbett, line 26: could dependabot do this for us?

gregcorbett (approved):
LGTM

.pre-commit-config.yaml
tofu-rocketry: I don't think so as this isn't a GitHub Action nor one of the other package ecosystems dependabot supports. Pre-commit's hosted CI can do this, but we are currently stuck with v2.5.0 for the other set of hooks while we're still supporting Python 2.

PR #332, opened 2023-10-24
Dbunloader refactor

tofu-rocketry:
Another refactoring PR in aid of the HEPscore work.

- Python 3 compatability fixes.
- Tidy of imports.
- Factoring out setting the number of records per message.
- Unit test for new factored out function.


tofu-rocketry:
Ah, need to fix how the logging is handled in the new function.

gregcorbett (approved):
LGTM

PR #333, opened 2023-10-25
Fix mysql tests

tofu-rocketry:
This solves the MySQL-based tests from hanging in CI when using Python 3.

The basic issue was the connections not being closed.


Sae126V:
LGTM.

Sae126V:
It's good to see that you have few asserts to what 3 supports.

Sae126V (approved):

PR #334, opened 2023-11-01
Verbosity fix

### LINKED ISSUE #324: Excessive parser logs from reporting unparsed files

RedProkofiev:
Minor change to @maxfischer2781 's commit, resolves #324  .  Adds a flag that checks if a file's been skipped, and sends a single non repeating message per scan_dir if this has occurred. 

Supersedes and closes #326.

### DIFF
--- a/bin/parser.py
+++ b/bin/parser.py
@@ -215,13 +215,13 @@ def scan_dir(parser, dirpath, reparse, expr, apel_db, processed):
                         updated.append(pr)
                 elif unparsed:
                     if not skipped_warning_flag:
-                        log.info("Files have been skipped: details can be found in debug.")
+                        log.info("Files skipped: rerun at DEBUG log level to see details.")
                         skipped_warning_flag = True
                     log.debug('Skipping file (failed to parse previously): %s',
                               abs_file)
                 else:
                     if not skipped_warning_flag:
-                        log.info("Files have been skipped: details can be found in debug.")
+                        log.info("Files skipped: rerun at DEBUG log level to see details.")
                         skipped_warning_flag = True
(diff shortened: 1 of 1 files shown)


tofu-rocketry:
Any particular reason for this solution compared with, say, doing a single `if skipped_warning_flag` check after the loop? Not saying either one is better, just interested in your thought process.

tofu-rocketry:
> My thought was that if for some reason a skippable parser fragment was the reason for a crash, and I didn't know it was going to debug, or if it preceded one, I'd want to the convenient message of "some output has gone to debug" before the crash whereas end of the loop it may never reach it.
> 
> It's just aesthetics though, if you prefer that I can change that now

No need; that's a good reason to have it this way which I hadn't though of.

tofu-rocketry (approved):

PR #335, opened 2023-11-08
Refactor load_from_msg

tofu-rocketry:
Another HEPscore supporting refactor.

- Change string splitting and stripping to a simpler list comprehension.
- Replace call to `set_all` to more direct `set_field`.
- Remove explicit string continuation as we can use implicit string continuation in parentheses.


tofu-rocketry:
Issues with Travis meant I needed to delete branch and re-push to invoke CI.

tofu-rocketry:
Coverage went down a smidge as I removed a covered line.

Sae126V (approved):
LGTM.  It's down to using x in the loop. You have used it because it is short lived variable and easily readable and accepted by PEP8 if I am NOT wrong.

RedProkofiev (approved):
Perfectly logical change and stylistically fine, let's ship it

PR #337, opened 2023-11-14
297 hepscore

tofu-rocketry:


apel/db/records/record.py
```diff
@@ -380,3 +430,22 @@ def _check_fields(self):
...
+                # Covert keys to lower case
+                cleaned_dict = {k.lower(): v for k, v in cleaned_dict.items()}
+            except ValueError as e:
+                raise InvalidRecordException("Expecting dictionary-like value. %s" % e)
+
+            for benchmark_type in benchmark_priority:
+                if benchmark_type in cleaned_dict:
+                    return benchmark_type, cleaned_dict[benchmark_type]
+            else:
+                raise InvalidRecordException("No valid benchmark type found")
```
RedProkofiev, line 451: I think these lines should be indented so that the else is under the purview of the if, not the for.  It's dead code right now.

apel/db/records/normalised_summary.py
```diff
@@ -62,8 +62,15 @@ def __init__(self):
...
-        self._db_fields = self._msg_fields
+        # This list specifies the information that goes in the database. It includes
+        # the additional ServiceLevelType field.
+        self._db_fields = ["Site", "Month", "Year", "GlobalUserName", "VO",
+                            "VOGroup", "VORole", "SubmitHost", "Infrastructure",
+                            "ServiceLevelType",
+                            "NodeCount", "Processors", "EarliestEndTime",
+                            "LatestEndTime", "WallDuration", "CpuDuration",
+                            "NormalisedWallDuration", "NormalisedCpuDuration",
+                            "NumberOfJobs"]
```
RedProkofiev, line 73: Is there a particular reason why in this circumstance we add ServiceLevelType in the middle of these fields?  Also, if we're just adding SLT and we're not changing any of the values in the fields after init, could we do this to minimise the risk of typing errors in the init field?
```self._db_fields = self._msg_fields.copy().insert("ServiceLevelType", position_n)```
Would just like to hear your commentary on this really :)

File: apel/db/records/summary.py [hunk 3]
```diff
@@ -225,3 +227,53 @@ def get_ur(self):
...
+        # Fields which will have an integer stored in them
+        self._int_fields = ["Month", "Year", "NodeCount", "Processors",
+                            "WallDuration", "CpuDuration", "NumberOfJobs"]
+
+        self._float_fields = ["ServiceLevel"]
+
+        self._datetime_fields = ["EarliestEndTime", "LatestEndTime"]
+
+        # Fields which should contain associative arrays in the message
+        self._dict_fields = ["ServiceLevel"]
```
RedProkofiev, line 279: Again, not an expert, should ServiceLevel be considered in two different fields for two different datatypes?  Is this one of those things where SL is going to be a float in one case and a dict in others?

test/test_job_record.py
```diff
@@ -98,5 +99,32 @@
...
+        record.set_field('WallDuration', 3600)
+        record.set_field('CpuDuration', 3600)
+        record.set_field('StartTime', 1234)
+        record.set_field('EndTime', 14234)
+        record.set_field('ServiceLevel', 10.3652782)
+        record.set_field('ServiceLevelType', 'HEPscore23')
+
+        try:
+            record._check_fields()
+        except:
```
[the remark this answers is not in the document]
RedProkofiev: Should handle InvalidRecordException :)

RedProkofiev (commented):
Looks like good changes, just a few minor quirks and me asking a few questions.  Also I think some commits should be squashed, from an outside perspective it's very hard to see a general picture of everything that's going on

apel/db/loader/record_factory.py
```diff
@@ -90,16 +93,27 @@ def create_records(self, msg_text):
...
-                    RecordFactory.SYNC_HEADER: SyncRecord,
-                    RecordFactory.CLOUD_HEADER: CloudRecord,
-                    RecordFactory.CLOUD_SUMMARY_HEADER: CloudSummaryRecord,
-                }
+                msg_header_version = header[index:].strip(': ')
+
+                if msg_header_version == "v0.4":
+                    # This routes us via the new sublcassed v0.4 record objects.
+                    record_mapping = {
+                        RecordFactory.JR_HEADER: JobRecord04,
```
Sae126V, line 101: Do we need to map JobRecord04 version to JR_04_HEADER something like that. Or is it we are referring to the same header and asking it to refer different classes in both cases?

apel/db/records/normalised_summary.py
```diff
@@ -234,3 +241,56 @@ def get_ur(self):
         # We don't want the XML declaration, because the whole XML
         # document will be assembled by another part of the program.
         return doc.documentElement.toxml()
+
+class NormalisedSummaryRecord04(NormalisedSummaryRecord):
+    """Class to represent a normalised summary record using the 0.4 message format
+
+    It differs from NormalisedSummaryRecord by adding a separate
+    ServiceLevelType field in the db fields, which is extracted from the
+    associative array in the Normalised...Duration fields before putting
```
Sae126V, line 250: Do we need spread syntax(...) here ?

apel/db/records/record.py
tofu-rocketry: Ah, yes, I think you're right.

test/test_job_record.py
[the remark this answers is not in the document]
tofu-rocketry: Think this was me being lazy and copying the `TestJobRecord` code.

apel/db/loader/record_factory.py
tofu-rocketry: This probably could be neater, but the problem is that we need to accept headers with versions < 0.4 as the original records, and ones with 0.4 as the new type, so we can't simply match on a complete header string.

apel/db/records/normalised_summary.py
tofu-rocketry: This list matches the order of the columns in the database, hence needing to be in the middle.

Doing it programmatically probably would be safer, though I'd want to make it clear where the column is occurring.

bin/dbunloader.py
```diff
@@ -158,7 +159,7 @@ def _bounded_records_per_message(config_object, logger):
 
     interval = cp.get('unloader', 'interval')
 
-    unloader = DbUnloader(db, unload_dir, include_vos, exclude_vos, local_jobs, withhold_dns)
+    unloader = DbUnloader(db, unload_dir, include_vos, exclude_vos, local_jobs, withhold_dns, dict_records)
```
Sae126V, line 162: I guess you might be having the changes ready already. It's about replicating the same in other places as well ? like client.py

apel/db/unloader.py
```diff
@@ -72,6 +91,7 @@ def _get_base_query(self, record_type):
         query = Query()
 
         if record_type in (JobRecord, SummaryRecord, NormalisedSummaryRecord,
+                           JobRecord04, SummaryRecord04, NormalisedSummaryRecord04,
```
Sae126V, line 94: Just a clarification: 

Say, Before if was a like JobRecord, NormalisedSummaryRecord we were doing the checks like whether they are having include_vos, exclude_vos or local_jobs etc..

Since we have added new record types(like JobRecord04, NormalisedSummaryRecord04)  Are we NOT doing the check anymore for the previous record types?

bin/dbunloader.py
tofu-rocketry: You mean in other places that need to unload?

Sae126V: like in client.py, When client has unloader enabled case?

scripts/update-x.y.z-2.1.0.sql
```diff
@@ -0,0 +1,152 @@
+-- This script contains a SQL block that can update
+-- APEL version x.x.x databases of the following types to 2.1.0:
```
Sae126V, line 2: Added for @Sae126V: Need to change x.x.x to previous release once finalized.

apel/db/records/summary.py
```diff
@@ -225,3 +227,53 @@ def get_ur(self):
         # We don't want the XML declaration, because the whole XML
         # document will be assembled by another part of the program.
         return doc.documentElement.toxml()
+
+
+class SummaryRecord04(SummaryRecord):
```
Sae126V, line 232: Since #359 got closed. Added for @Sae126V . Should ensure to leverage the base class which we are extending.

~ Similarly for JobRecord04, NormalisedSummaryRecord04 class.

apel/db/records/summary.py
tofu-rocketry: Depending on the accounting record version.

apel/db/records/record.py
tofu-rocketry: Ah, no, this is to raise an exception is the for-loop is exhausted without finding a valid benchmark type.

apel/db/unloader.py
```diff
@@ -61,10 +64,22 @@ class DbUnloader(object):
...
+                         SummaryRecord04: SUMMARY_MSG_HEADER_04,
+                         NormalisedSummaryRecord04: NORMALISED_SUMMARY_MSG_HEADER_04}
+
+
     # all record types for which withholding DNs is a valid option
     MAY_WITHHOLD_DNS = [JobRecord, SyncRecord, CloudRecord]
 
-    def __init__(self, db, qpath, inc_vos=None, exc_vos=None, local=False, withhold_dns=False):
+    def __init__(self, db, qpath, inc_vos=None, exc_vos=None, local=False, withhold_dns=False,
+                 dict_records=False):
```
Sae126V, line 82: Just a clarification. Since `dict_records ` is NOT saved here for later use.  Is it intentional, Adrian ? Would that be problem later on?
like either public `self.dict_records = dict_records`  or private.

bin/dbunloader.py
Sae126V: Just a clarification, Adrian. In bin/client.py -> We have DbUnloader. If we want to set or use 'dict_records=True' case it is NOT possible via `client.py` as of now. Is this expected ?

Sae126V:
Finished my review. This PR LGTM!

- Once I get the response for [this](https://github.com/apel/apel/pull/337#discussion_r1965288928). I am happy to approve this

bin/dbunloader.py
tofu-rocketry: Added in latest commit.

Sae126V (approved):
LGTM!

PR #342, opened 2023-12-20
Fix HTCondor format for trailing empty fields

### LINKED ISSUE #341: HTCondor parser adds empty field for common/documented HTCondor queries
The parser and [example input](https://github.com/apel/apel/blob/9d0e6e8269b56b0e8e478d6e37cd9dcb19d08738/apel/parsers/htcondor.py#L42) assume a format of the form `a|b|...|y|z` with `|` as a value *delimiter*. In contrast, both the [example query](https://github.com/apel/apel/blob/9d0e6e8269b56b0e8e478d6e37cd9dcb19d08738/apel/parsers/htcondor.py#L41) as well as [previous](https://github.com/htcondor/htcondor-ce/blob/37451e88abfb04f76194a96a89a20f88e78a0060/contrib/apelscripts/condor_batch.sh#L30) and [current](https://github.com/htcondor/htcondor-ce/blob/8145b5c6c92275b5c72cc0b28f01b197681886f4/contrib/apelscripts/condor_batch_blah.py#L159) HTCondor-CE produce the form `a|b|...|y|z|` with `|` as value *terminator*. This means that APEL parses the actual format always with a trailing, empty string which fails [the cpu scaling conversion](https://github.com/apel/apel/blob/9d0e6e8269b56b0e8e478d6e37cd9dcb19d08738/apel/parsers/htcondor.py#L49).  
This likely didn't lead to major issues as the default for producers is to include the scale factor and APEL ignores everything past the 11th field. However, it means the parser is not able to parse records without scaling factor.

-----------

This ticket is for reference. I'm currently preparing a PR to fix the issue.

maxfischer2781:
This PR adjusts the `htcondor` parser to match its own example and HTCondor-CE. It is still compatible with the old format.

- HTCondor parser now ignores a trailing, *empty* field after all required ones. This means the final `|` can act both as a delimiter and terminator.
- The example is adjusted for consistency: The query now has a newline to make it feasible for multiple jobs, and the output has the trailing `|` of the query.

Closes #341.

### DIFF
--- a/apel/parsers/htcondor.py
+++ b/apel/parsers/htcondor.py
@@ -39,7 +39,7 @@ class HTCondorParser(Parser):
         Parses single line from accounting log file.
         '''
         # condor_history -constraint "JobStartDate > 0" -format "%s|" GlobalJobId -format "%s|" Owner -format "%d|" RemoteWallClockTime -format "%d|" RemoteUserCpu -format "%d|" RemoteSysCpu -format "%d|" JobStartDate -format "%d|" EnteredCurrentStatus -format "%d|" ResidentSetSize_RAW -format "%d|" ImageSize_RAW -format "%d|\n" RequestCpus
-        # arcce.rl.ac.uk#2376.0#71589|tatls011|287|107|11|1435671643|1435671930|26636|26832|1|1
+        # arcce.rl.ac.uk#2376.0#71589|tatls011|287|107|11|1435671643|1435671930|26636|26832|1|1|
 
         values = line.strip().split('|')
 
--- a/apel/parsers/htcondor.py
+++ b/apel/parsers/htcondor.py
@@ -45,7 +45,7 @@ class HTCondorParser(Parser):
 
         # Set scaling factor using value from log if appended to log line.
         cputmult = float(1.0)
-        if len(values) > 10:
+        if len(values) > 10 and values[10]:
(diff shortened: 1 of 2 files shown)


maxfischer2781:
Ugh, the Codacy lint now complains about the line length but that was already too long before. Breaking up the line to fit in the length seems counterproductive, since reproducing the actual format is kind of the point here.

What's the policy here? Should I break up the line regardless? If I'm reading the Codacy report right, this is just prettified Pylint report - is it okay to [disable the inspection for those lines or the entire block](https://pylint.readthedocs.io/en/latest/user_guide/messages/message_control.html#block-disables)?

tofu-rocketry:
Don't worry about the line length. I need to sort the code style checks to have a more generous limit, but that requires a general overhaul of the CI tools.

maxfischer2781:
@tofu-rocketry Thanks for updating the style checks, this is ready for review.

tofu-rocketry (approved):

tofu-rocketry (changes requested):
Oops, realised I should update the unit tests. Can sort that myself.

tofu-rocketry (approved):

gregcorbett (approved):
LGTM

tofu-rocketry:
Thanks, @gregcorbett

PR #348, opened 2024-01-29
Add accelerator records and summaries (modifications for CASO and formatting)

nick-whyatt:
Editing the above PR to work with CASO by changing some string formatting and tests.


schemas/accelerator.sql
```diff
@@ -0,0 +1,397 @@
+-- This schema adds the tables necessary for Accelerator accounting as a
+-- separate record as part of the wider Cloud Accounting system.
```
tofu-rocketry, line 2: It's not dependant on `cloud.sql` though, is it?

requirements.txt
```diff
@@ -4,3 +4,4 @@ MySQL-python
 python-ldap
 iso8601
 dirq
+jsonschema
```
tofu-rocketry, line 7: Can you find the relevant rpm to add to the SPEC file too?

apel/common/json_utils.py
```diff
@@ -0,0 +1,29 @@
...
+    "Type": "%s",
+    "Version": "%s",
+    "UsageRecords": %s
+}"""
+
+def to_message(type, version, *usage_records):
+    """ Generate JSON message format """
+    return (
+        JSON_MSG_BOILERPLATE % (type, version, list(usage_records))
+    ).replace("'", '"')
```
tofu-rocketry, line 29: Needs a newline at the end of the file.

apel/db/records/accelerator.py
```diff
@@ -0,0 +1,85 @@
...
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+
+"""
+
+from apel.common import parse_fqan
```
tofu-rocketry, line 18: Unused.

apel/db/records/accelerator_summary.py
```diff
@@ -0,0 +1,84 @@
...
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+
+"""
+
+from apel.common import parse_fqan
```
tofu-rocketry, line 18: Unused.

apel/common/json_utils.py
```diff
@@ -0,0 +1,29 @@
...
+Helper methods to perform JSON operations.
+"""
+
+JSON_MSG_BOILERPLATE = """{
+    "Type": "%s",
+    "Version": "%s",
+    "UsageRecords": %s
+}"""
+
+def to_message(type, version, *usage_records):
```
tofu-rocketry, line 25: Arg name needs changing - `type` is a Python built-in.

schemas/accelerator.sql
RedProkofiev: It is

RedProkofiev: Just had a look to confirm - the only coupling is that one of the functions (SiteLookup()) is used in ReplaceAcceleratorRecord and that function is contained within cloud.sql.  I think the best approach here would be to set an issue to decouple all the miscellaneous lookup functions for categories and elements that are the same into a utils.sql, then run that inside accelerator.sql and/or cloud.sql as there's currently some duplication.  It shouldn't be a hard fix but I don't have time to figure it out

tofu-rocketry:
> this is not a fun conflict :(

I've figured out the rebase. You didn't touch anything that had actually been changed in `dev`, but the changes are just adjacent enough that they conflicted.

tofu-rocketry (commented):
From Greg on #264:
> `SiteName VARCHAR(255) NOT NULL` should be `SiteID INT NOT NULL` to be consistent with other forms of APEL accounting (both in the `AcceleratorRecords` and `AcceleratorSummaries` tables) and various procedures updating accordingly.
> 
> Was the intention the accelerators would be an "add-on" to cloud or grid accounting, or its own standalone type of accounting? If it needs to be standalone - a `Sites` table and `SitesLookup` function would need to be added similar to https://github.com/apel/apel/blob/1.9.2-1/schemas/server.sql#L420

Yes, we will tweak the schema to make the accelerator accounting standalone as this will make it much easier to tweak and develop. It should also match the cloud schema a lot closer generally such as having update time columns.

tofu-rocketry:
Closed in favour of #459

PR #350, opened 2024-02-28
[REVIEW IS NOT NEEDED] Python3 compatible code

Sae126V:
This PR will be closed soon since #354 have raised breaking this down to multiple commits and NOT disturbing the flow of python2.7.

### DIFF
--- a/apel/common/hashing.py
+++ b/apel/common/hashing.py
@@ -14,10 +14,7 @@
    limitations under the License.
 '''
 # hashlib only available in python2.5+
-try:
-    from hashlib import md5
-except ImportError:
-    from md5 import md5
+from hashlib import md5
 import gzip
 
 def calculate_hash(fname):
@@ -29,21 +26,21 @@ def calculate_hash(fname):
     For sample usage please go to: apel2/bin/client.py
     '''
 
-    data = 'initial'
+    data = b'initial'
 
     md = md5()
 
     # try opening as a gzip file, and if it fails
     # try as a regular file
     try:
-        fp = gzip.open(fname, 'r')
-        while data != '':
+        fp = gzip.open(fname, 'rb')
+        while data != b'':
             # 128kiB buffer
             data = fp.read(131072)
             md.update(data)
-    except IOError: # not a gzipped file
-        fp = open(fname, 'r')
-        while data != '':
+    except (IOError, OSError): # not a gzipped file
+        fp = open(fname, 'rb')
+        while data != b'':
             # 128kiB buffer
             data = fp.read(131072)
             md.update(data)
--- a/apel/db/loader/loader.py
+++ b/apel/db/loader/loader.py
@@ -20,14 +20,14 @@ Module containing the Loader class.
 
 import logging
 import os
-from xml.parsers.expat import ExpatError
+from xml.parsers.expat import ExpatError, errors
 
 from dirq.queue import Queue
 
 import apel.db
 from apel.db.loader.xml_parser import XMLParserException
 from apel.db.records import InvalidRecordException
-from record_factory import RecordFactory, RecordFactoryException
+from .record_factory import RecordFactory, RecordFactoryException
 
 
 # set up the logger
@@ -114,7 +114,7 @@ class Loader(object):
                 os.remove(pidfile)
             else:
                 log.warning("pidfile %s not found.", pidfile)
-        except IOError as e:
+        except (IOError, OSError) as e:
             log.warning("Failed to remove pidfile %s: %s", pidfile, e)
             log.warning("The loader may not start again until it is removed.")
 
@@ -136,7 +136,7 @@ class Loader(object):
         while self.current_msg:
             if not self._inq.lock(self.current_msg):
                 log.warning("Skipping locked message %s", self.current_msg)
-                self.current_msg = self._inq.next()
+                self.current_msg = next(self._inq)
                 continue
             log.debug("Reading message %s", self.current_msg)
             data = self._inq.get(self.current_msg)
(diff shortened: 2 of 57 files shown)


apel/common/hashing.py
```diff
@@ -29,21 +26,21 @@
...
+        fp = gzip.open(fname, 'rb')
+        while data != b'':
             # 128kiB buffer
             data = fp.read(131072)
             md.update(data)
-    except IOError: # not a gzipped file
-        fp = open(fname, 'r')
-        while data != '':
+    except (IOError, OSError): # not a gzipped file
+        fp = open(fname, 'rb')
```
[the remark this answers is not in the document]
gregcorbett: could the file be opened with `with` here to ensure the file is always closed?

apel/parsers/lsf.py
```diff
@@ -126,9 +126,10 @@ def parse(self, line):
             data[key] = mapping[key](items)
 
         # Input checking
-        if data['CpuDuration'] < 0:
+        # TODO: Check needed, across codebase: When testing try to compare with schema
```
gregcorbett, line 129: Open an issue rather than add a TODO to the codebase.

Sae126V: I could have given a heads-up. TODO is the activity for my testing with rocky8 and 9 ticket. I will fix those as part of that. I can assure it is not major one. It is doable in very less time, Greg.  I will remove those eventually. Make sense ?

bin/retrieve_dns.py
```diff
@@ -114,14 +114,17 @@ def get_xml(url, proxy):
     '''
     try:
         # Try without a proxy
-        conn = urllib.urlopen(url)
+        conn = urllib.request.urlopen(url)
         dn_xml = conn.read()
         conn.close()
-    except IOError:
+        # TODO: If we are the owners of the url specified then we can catch proper errors than all.
+    except Exception:
```
gregcorbett, line 121: We probably want to catch more specific exceptions here?

apel/db/loader/loader.py
```diff
@@ -136,7 +136,7 @@ def load_all_msgs(self):
         while self.current_msg:
             if not self._inq.lock(self.current_msg):
                 log.warning("Skipping locked message %s", self.current_msg)
-                self.current_msg = self._inq.next()
+                self.current_msg = next(self._inq)
```
gregcorbett, line 139: ah okay, this is the Python 3 way of using next - see: https://peps.python.org/pep-3114/

apel/common/hashing.py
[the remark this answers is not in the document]
gregcorbett: but also, not an issue you've introduced so 🤷

apel/db/loader/loader.py
gregcorbett: though would this break python2.7 support? Does `.next()` still exist in Python 3?

Maybe we need:

```
try:
    self.current_msg = next(self._inq)
except SomeException:
    self.current_msg = self._inq.next()
```

bin/retrieve_dns.py
gregcorbett: `URLError` reads to me as the base class for all `urllib.request` exceptions: https://docs.python.org/3/library/urllib.error.html#urllib.error.URLError, which would be better to catch then just `Exception`.

> The [urllib.error](https://docs.python.org/3/library/urllib.error.html#module-urllib.error) module defines the exception classes for exceptions raised by [urllib.request](https://docs.python.org/3/library/urllib.request.html#module-urllib.request). The base exception class is [URLError](https://docs.python.org/3/library/urllib.error.html#urllib.error.URLError).

tofu-rocketry:
These are a lot of changes squashed into one commit. It'd be happier if the changes were grouped by category into separate commits.

I thought I'd pointed you at the start I'd made here: https://github.com/apel/apel/compare/dev...tofu-rocketry:apel:python3

Note the separate commits for groups of changes. e.g. separate commits for "Change implicit relative imports to explicit" and "Change .iteritems() to .items() as removed in Py3"

Sae126V:
Sure Adrian. Right from dev activity start, I have not realized I have to do the same breaking into multiple commits. But, I am happy to try and break it into smaller commits.

bin/client.py
```diff
@@ -31,11 +31,7 @@
 import logging.config
 import ldap
 
-try:
-    import ConfigParser
-except ImportError:
-    # Renamed in Python 3
-    import configparser as ConfigParser
+import configparser
```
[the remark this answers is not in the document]
Sae126V: Now that we are not disturbing the flow. I have added this in #354  5a6b0c17a65f9d0b363b5b86e3afcbde9a08d5cf. Just to acknowledge, Greg.

bin/retrieve_dns.py
Sae126V: I have added this in https://github.com/apel/apel/pull/354 https://github.com/apel/apel/commit/5b581a25e47958f5cd9a867dcb80609d4ab72d82. Just to acknowledge, Greg

apel/db/loader/loader.py
Sae126V: I have added this in https://github.com/apel/apel/pull/354 https://github.com/apel/apel/commit/ef8d4b3312aee2ac17524ad1c566ff3856d64be3 and dd4bd0d3fe02d7a86c33f84c9f8beb117f8d2fa7.

This next() will be handled by future for cross-compatibility(dd4bd0d3fe02d7a86c33f84c9f8beb117f8d2fa7)

tofu-rocketry:
Closing in favour of #354

PR #352, opened 2024-03-08
Update and simplify upload-artifact action version

tofu-rocketry:
Setting the version to just major.minor rather than major.minor.patch and letting the Action pull in the latest of the minor series should hopefully reduce the number of dependabot PRs a little while still giving us a good view on relevant changes.

Inspired by https://github.com/GOCDB/gocdb-failover-scripts/pull/30

And reminded by: https://github.com/apel/apel/pull/349


RedProkofiev (approved):
I saw this process with Greg's earlier commit on a different repository so if it works as intended no reason not to approve!

PR #353, opened 2024-03-11
GitHub action for running unit tests

Sae126V:
This is specific to 3.6 or 3.9 versions. And verified GitHub action runs successfully 

If we want to make it generic. we can make it as 3.x.

Note:
This is having dependency on the APEL codebase (Codebase should be in py3.x)

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -1,0 +1,59 @@
+name: Run Unit Tests
+
+on: [push, pull_request]
+
+jobs:
+  unit-test:
+    runs-on: ubuntu-20.04
+    services:
+      mariadb:
+        image: mariadb:10.4
+        env:
+          MARIADB_ALLOW_EMPTY_ROOT_PASSWORD: 'yes'
+          MARIADB_DATABASE: apel_unittest
+        ports:
+          - 3306:3306
+        options: >-
+          --health-cmd="mysqladmin ping -h localhost"
+          --health-interval=10s
+          --health-timeout=5s
+          --health-retries=3
+
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ['3.6', '3.9']
+    name: Python ${{ matrix.python-version }} test
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Install system dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y build-essential python3-dev libldap2-dev libsasl2-dev libmysqlclient-dev
+
+    - name: Install Python dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
+        pip install -r requirements-test.txt
+
+    - name: Pre-test setup
+      run: |
+        export TMPDIR=$PWD/tmp
+        mkdir -p $TMPDIR
+        export PYTHONPATH=$PYTHONPATH:`pwd -P`
+        cd test
+
+    - name: Run unit tests
+      run: coverage run --branch --source=apel,bin -m unittest discover -s test -p 'test_*.py'
+      working-directory: ${{ github.workspace }}
+
+    - name: Upload coverage to Codecov
+      uses: codecov/codecov-action@v4
+  
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -13,6 +13,15 @@ jobs:
           MARIADB_DATABASE: apel_unittest
         ports:
           - 3306:3306
+        ################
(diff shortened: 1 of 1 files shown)


.github/workflows/unit-test.yml
```diff
@@ -0,0 +1,59 @@
...
+    runs-on: ubuntu-20.04
+    services:
+      mariadb:
+        image: mariadb:10.4
+        env:
+          MARIADB_ALLOW_EMPTY_ROOT_PASSWORD: 'yes'
+          MARIADB_DATABASE: apel_unittest
+        ports:
+          - 3306:3306
+        options: >-
```
tofu-rocketry, line 16: Can you add some comments to explain these options, please?

.github/workflows/unit-test.yml
```diff
@@ -0,0 +1,59 @@
...
+
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Install system dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y build-essential python3-dev libldap2-dev libsasl2-dev libmysqlclient-dev
```
tofu-rocketry, line 38: What are these actually needed for? The name of this section doesn't seem to describe the right thing.

.github/workflows/unit-test.yml
```diff
@@ -0,0 +1,59 @@
...
+
+    - name: Install Python dependencies
```
tofu-rocketry, line 40: Could you break this up like SSM does into separate sections. That means that if one of them fails the errors will be clearer.

.github/workflows/unit-test.yml
```diff
@@ -0,0 +1,59 @@
...
+      run: |
+        python -m pip install --upgrade pip
```
tofu-rocketry, line 42: Is a pip upgrade really needed?

.github/workflows/unit-test.yml
```diff
@@ -0,0 +1,59 @@
...
+
+    - name: Pre-test setup
+      run: |
+        export TMPDIR=$PWD/tmp
+        mkdir -p $TMPDIR
+        export PYTHONPATH=$PYTHONPATH:`pwd -P`
+        cd test
+
+    - name: Run unit tests
+      run: coverage run --branch --source=apel,bin -m unittest discover -s test -p 'test_*.py'
```
tofu-rocketry, line 54: Shouldn't need to add `-s test -p 'test_*.py'`

tofu-rocketry: `--buffer` would be a useful option to add.

tofu-rocketry (changes requested):
Can you fix the pre-commit issues too, please?

.github/workflows/unit-test.yml
Sae126V: For some reason I couldn't run with the command below. So, I have to specify `-s test` at the least to successfully run the testcases.

Not working in my case:
`coverage run --branch --source=apel,bin -m unittest discover --buffer`


Working if I explicitly specify the -s indicating the start-directory.
`coverage run --branch --source=apel,bin -m unittest discover -s test --buffer`

.github/workflows/unit-test.yml
Sae126V: Sure. I have changed the name and added relevant dependencies to the libraries we are using.

.github/workflows/unit-test.yml
Sae126V: Please see the comments below https://github.com/apel/apel/pull/353#discussion_r1530763061 , Adrian.

.github/workflows/unit-test.yml
Sae126V: These options are docker container options to see the container health status. Not a dependency thou to successfully run the testcases. But, if we provide health checks, the container will not start until the db container is healthy (basically until mysqladmin is up and accepting connections).

mysqladmin ping will return a false positive if the server is running but not yet accepting connections.


Just to summarise from the link below:

`--health-cmd	        Command to run to check health `
`--health-interval	Time between running the check`
`--health-retries	Consecutive failures needed to report unhealthy`
`--health-timeout	Maximum time to allow one check to run\n`



For further information. Please refer to https://docs.docker.com/engine/reference/run/#healthchecks

tofu-rocketry: Can you add the summarised comments to the code please?

tofu-rocketry:
Can you fix the pre-commit whitespace issues too, please?

.github/workflows/unit-test.yml
Sae126V: Sure Adrian. Added this comments into the code now. [69c7119](https://github.com/apel/apel/pull/353/commits/69c71195d08fb9420eef3358e50f21fe4cadb263)

tofu-rocketry (approved):

PR #354, opened 2024-03-18
Apel codebase migration to python3

Sae126V:
Resolves GT-265

This is a clean PR with smaller commits on top of #350 .

Few Important ones. Thought to share with you all.

[de09670] [Add future for backward compatibility](https://github.com/apel/apel/pull/354/commits/de09670f710ff546b6c40ac8e7e4202ee41170e3)
Added it so that we can run Python 3 code and to run it with minimal effort under Python 2. We can also run it with python 3.

Ref: https://pypi.org/project/future/0.6.0/
___________

[[ecaaef0](https://github.com/apel/apel/pull/354/commits/ecaaef080801cc61277d2791cd23c4fad41535b6)] [Add mysqlclient to support py3.x](https://github.com/apel/apel/pull/354/commits/ecaaef080801cc61277d2791cd23c4fad41535b6)
**Make sure to run and install mysqlclient with sudo**

This project is a fork of [MySQLdb1](https://github.com/farcepest/MySQLdb1). This project adds Python 3 support.

Ref: https://mysqlclient.readthedocs.io/
___________

[[5a6b0c1](https://github.com/apel/apel/pull/354/commits/5a6b0c17a65f9d0b363b5b86e3afcbde9a08d5cf)] Add try and except block for configparser to work in py3   

Since configparser ONLY works in python3+. Added try and except block.

Python2.7: https://docs.python.org/2.7/library/configparser.html
Python3.6: https://docs.python.org/3.6/library/configparser.html
__________
[5b581a2](https://github.com/apel/apel/pull/354/commits/5b581a25e47958f5cd9a867dcb80609d4ab72d82)  Add new method to fix urllib issues with py3 compatibility
`urllib` went significant change in the code flow. Added another method to support python2.7 and python3.6+ explicitly.

Python2.7: https://docs.python.org/2.7/library/urllib.html
Python 3: https://docs.python.org/3.6/library/urllib.html
__________

[429fdc7](https://github.com/apel/apel/pull/354/commits/429fdc78e612c7b60ed470869d9caaf59e9c41e4)  Update code to handle flow for expat errors

Since the return type got changed. Separate flow for the python is needed.
Ref: https://docs.python.org/3.6/library/pyexpat.html
Ref: https://docs.python.org/2.7/library/pyexpat.html

__________
[2f33a1e](https://github.com/apel/apel/pull/354/commits/2f33a1e9c8f9f4ea6e28dffab5411b84b6634cf1) Update code to fix the data type (.write) for py2&3 compatibility

Strings are bytes by default in python 2.7. Need to convert strings to bytes which can then be passed to os.write(). 
Python2.7: https://docs.python.org/2.7/library/os.html
Python3.6: https://docs.python.org/3.6/library/os.html
__________

General fixes as part of python3 migration:

Ref: https://docs.python.org/3.1/whatsnew/3.0.html
Ref: https://python-future.org/

### DIFF
--- a/test/test_mysql.py
+++ b/test/test_mysql.py
@@ -121,6 +121,10 @@ class MysqlTest(unittest.TestCase):
         cloud4_mf = apel.db.records.cloud.CloudRecord()
         cloud4_mf.load_from_msg(CLOUD4_MISSING_FIELDS)
 
+        # items_in = list(cloud2._record_content.items())
+        # items_in += list(cloud4._record_content.items())
+        # items_in += list(cloud4_nb._record_content.items())
+        # items_in += list(cloud4_mf._record_content.items())
 
         record_list = [cloud2, cloud4, cloud4_nb, cloud4_mf]
 


Sae126V:
Dropped Commit ID `68eccae`. As it NOT needed because OSError is an alias for IOError.

Sae126V:
Added fbbd8a3de391af497f96d84cc1018def150d2577. It was missed from fc7c2d4c5dd6fe9fd43753ac5e0f048fce8ba4a6 `loader.py`.

requirements.txt
```diff
@@ -1,6 +1,7 @@
 # Base requirements for apel
 
-MySQL-python
-python-ldap<3.4.0  # python-ldap-3.4.0 dropped support for Python 2
-iso8601
+mysqlclient==2.1.1
+python-ldap
+iso8601==1.1.0
 dirq
+future
```
tofu-rocketry, line 7: Can you add a comment that this provides `future`. Is there an rpm package easily available that can be added to the SPEC file?

constraints.txt
```diff
@@ -1,5 +0,0 @@
-# Constraints that apply to pip installed requirements
```
tofu-rocketry, line 1: Why remove the constraints file?

scripts/msg_status.py
```diff
@@ -19,7 +19,9 @@
     are locked, it gives the option to remove the lock.
     @author: Will Rogers
 '''
+from __future__ import print_function
 
+from builtins import next
```
tofu-rocketry, line 24: Where there's import changes, then we should try and follow [PEP8 ordering](https://peps.python.org/pep-0008/#imports), at least for the grouping. So `builtins` should go after `os` (standard lib) and just before `dirq` (3rd-party). There will be other places to pay attention to this.

setup.py
```diff
@@ -63,6 +63,12 @@ def main():
     log_rotate_dir = '/etc/logrotate.d'
     log_rotate_files = ['scripts/apel-client']
 
+    # Python 2.x dependencies
+    _install_requires = ['MySQL-python', 'iso8601', 'python-ldap', 'dirq', 'future']
```
tofu-rocketry, line 67: Can you flip this so that Python 3 is the default set and Python 2 modifies it? Then it'll be easier to remove in future.

test/test_blah.py
```diff
@@ -1,3 +1,4 @@
+from builtins import zip
```
tofu-rocketry, line 1: 3rd party import after standard lib.

test/test_mysql.py
```diff
@@ -44,7 +44,7 @@ def test_test_connection(self):
     def test_bad_connection(self):
         """Check that initialising ApelDb fails if a bad password is used."""
         self.assertRaises(apel.db.apeldb.ApelDbException, apel.db.apeldb.ApelDb,
-                          'mysql', 'localhost', 3306, 'root', 'badpassword',
```
tofu-rocketry, line 47: Does "localhost" not work?

tofu-rocketry (changes requested):

tofu-rocketry:
When asked to rebase, please don't do merges do update the code as it's not the same thing. I can talk you through the process if you need it clarified.
image

requirements.txt
```diff
@@ -1,6 +1,7 @@
 # Base requirements for apel
 
-MySQL-python
-python-ldap<3.4.0  # python-ldap-3.4.0 dropped support for Python 2
-iso8601
+mysqlclient==2.1.1
```
tofu-rocketry, line 3: Why this version? Specifying anything later than 1.4.6 breaks the Travis Python 2 build.

requirements.txt
```diff
@@ -1,6 +1,7 @@
...
+python-ldap
+iso8601==1.1.0
```
tofu-rocketry, line 5: Why this version?

requirements-test.txt
```diff
@@ -1,10 +1,10 @@
 # Additional requirements for the unit and coverage tests
 
-# Constraints on the requirements below
--c constraints.txt
+pycparser
+idna
 
 unittest2
-coveralls<=1.2.0
-mock<4.0.0  # Pinned because version 4 dropped support for Python 2.7
```
tofu-rocketry, line 8: Removing the constraint is likely to break the Travis Python 2 build (which the requirements files are mostly used for).

requirements.txt
Sae126V: Why this version?
-> Acc. to [2.1.1 ](https://pypi.org/project/mysqlclient/2.1.1/). It acknowledges that it works with 3.6 and 3.9 which we tested the codebase and this is the latest package. From 2.2.0 they dropped support for py3.6 officially.

Specifying anything later than 1.4.6 breaks the Travis Python 2 build.
-> Cool. I have added the separate flow for Travis now it should work as expected. Aimed to test codebase with py2 and haven't disturbed the previous libs with version info limitations.

requirements.txt
Sae126V: Same as above. Acc. to 1.1.0 https://pypi.org/project/iso8601/1.1.0/   . It acknowledges that it works with 3.6.2+ and 3.9 which we tested the codebase and this is the latest package. Anything above that requires 3.7 codebase.

requirements-test.txt
Sae126V: (Same respone as before) I have added the separate flow for Travis now it should work as expected. Aimed to test codebase with py2 and haven't disturbed the previous libs with version info limitations.

setup.py
Sae126V: Sure. Make sense added in 45852290173e7bd4e79ea067e58efe3de4454d6c

constraints.txt
Sae126V: I was aiming to make it work and TEST codebase with 3.6 and 3.9. Since we don't need explicit file with limitations naming `constraints.txt` because it works without those explicit limitations of libs mentioned in constraints.txt file . Thus, I have added it to the `requirements-test.txt`  file directly.

But, I understood we don't want to disturb the flow of that as of now. I have added a separate flow in  https://github.com/apel/apel/commit/09d1e62db2531ad565deb4ea2f95486e56e908b6.

requirements.txt
Sae126V: Sure. Added comments for future explaining why we are using this in short now. And also added the relevant rpm in SPEC file to make it work with RPM commands itself.

scripts/msg_status.py
Sae126V: According to https://python-future.org/imports.html and https://python-future.org/overview.html. 

It was meant to be at the top of the file. Seems it is a Known pollution to make the code work in py2.

I would keep this in the same position and corrected the other position and differentiate the standard imports to the future relevant changes at the first.- Just to highlight in a31fe35e089d9960ae60d34a2107c44717747221

apel/parsers/sge.py
```diff
@@ -89,9 +89,9 @@ def _load_multipliers(self):
         hosts = xml_str.getElementsByTagName("host")
 
         for host in hosts:
-            host_name = dict(host.attributes.items())["name"]
+            host_name = dict(list(host.attributes.items()))["name"]
             for resource in host.getElementsByTagName("resourcevalue"):
-                resource_name = dict(resource.attributes.items())["name"]
+                resource_name = dict(list(resource.attributes.items()))["name"]
```
maxfischer2781, line 94: Since the list is immediately consumed by dict, there is no advantage to the list. Just remove it.

That said, it seems superfluous to construct an entire `dict` only to access one item *by key*. You should be able to do something like `host.attributes["name"]` (or similar - Python 2 may need an explicitly named method).

test/test_car_parser.py
```diff
@@ -178,7 +178,7 @@ def test_car_parser(self):
             self.assertTrue('StartTime' in cont)
             self.assertTrue('EndTime' in cont)
 
-            for key in cases[car].keys():
+            for key in list(cases[car].keys()):
```
maxfischer2781, line 181: Superfluous `list(….keys()` here.
```suggestion
            for key in cases[car]:
```

test/test_htcondor_parser.py
```diff
@@ -53,20 +55,20 @@ def test_parse_line_no_cputmult(self):
 
         cases = {}
         for line, value in zip(lines, values):
-            cases[line] = dict(zip(keys, value))
+            cases[line] = dict(list(zip(keys, value)))
```
maxfischer2781, line 58: The `list` (and the `.key()`) calls in this file changeset are superfluous. See previous comments for explanations.

maxfischer2781 (commented):
Good to see this making progress! I couldn't help but go through the changeset and add some suggestions along the way; mostly superfluous refactoring (e.g. intermediate `list`s when that is not needed) but also some dated code that can be "modernised" on this opportunitiy.

Feel free to disregard the suggestions if they aren't wanted.

Sae126V:
Firstly, Thank you @maxfischer2781. Much appreciated!

Well, The approach taken earlier was NOT to touch the previous code(Unless agreed) and making it to work with py3.6 and py3.9. (For instance, same goes for opening and closing the file using `with` rather than the `for` etc... ) Thus, made the changes accordingly. But, I definitely agree with the point of using and considering complexity of the loop and the points mentioned are worth including to modernize the code. As stated earlier, they are definitely worth including, I will have a word with @tofu-rocketry and @gregcorbett on how to proceed.

test/test_htcondor_parser.py
Sae126V: Make sense @maxfischer2781. Added this in [daecf80](https://github.com/apel/apel/pull/354/commits/daecf8087f5404c800245f05929f65f1aa0800ab)

apel/parsers/pbs.py
```diff
@@ -118,7 +120,7 @@ def _parse_mpi(exec_host):
 
     core_info = exec_host.split('+')
     # Split hostname and core_no into seperate lists.
-    hostnames, core_no = zip(*[x.split('/') for x in core_info])
+    hostnames, core_no = list(zip(*[x.split('/') for x in core_info]))
```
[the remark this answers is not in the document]
Sae126V: Added this in [daecf80](https://github.com/apel/apel/pull/354/commits/daecf8087f5404c800245f05929f65f1aa0800ab)

apel/db/records/record.py
```diff
@@ -181,7 +185,7 @@ def load_from_tuple(self, tup):
         Given a tuple from a mysql database, load fields.
         '''
         assert len(tup) == len(self._db_fields), 'Different length of tuple and fields list'
-        self.set_all(dict(zip(self._db_fields, tup)))
+        self.set_all(dict(list(zip(self._db_fields, tup))))
```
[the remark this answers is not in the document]
Sae126V: Cool. Added this in [daecf80](https://github.com/apel/apel/pull/354/commits/daecf8087f5404c800245f05929f65f1aa0800ab)

apel/db/backends/oracle.py
```diff
@@ -85,13 +88,13 @@ def load_records(self, record_list, source):
                 cur.callproc(proc, values)
 
             # Now merge the temporary tables into the actual tables.
-            for k, v in self.MERGE_PROCEDURES.iteritems():
+            for v in list(self.MERGE_PROCEDURES.items()):
```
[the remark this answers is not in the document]
Sae126V: Added this in [daecf80](https://github.com/apel/apel/pull/354/commits/daecf8087f5404c800245f05929f65f1aa0800ab)

requirements-test.txt
```diff
@@ -1,10 +1,10 @@
 # Additional requirements for the unit and coverage tests
 
-# Constraints on the requirements below
--c constraints.txt
+pycparser
+idna
```
tofu-rocketry, line 4: Remove as we don't need to require (they're dependencies of our dependencies).

apel/db/loader/loader.py
```diff
@@ -157,7 +162,14 @@ def load_all_msgs(self):
             except (RecordFactoryException, LoaderException,
                     InvalidRecordException, apel.db.ApelDbException,
                     XMLParserException, ExpatError) as err:
-                errmsg = "Parsing unsuccessful: %s" % str(err)
+                if sys.version_info >= (3,):
+                    if isinstance(err, ExpatError):
```
tofu-rocketry, line 166: use `and`

scripts/msg_status.py
tofu-rocketry: Yeah, that's ok. As discussed, it's a bit like the `__future__` imports.

setup.py
```diff
@@ -63,6 +63,12 @@ def main():
     log_rotate_dir = '/etc/logrotate.d'
     log_rotate_files = ['scripts/apel-client']
 
+    # Python 3.x dependencies
+    _install_requires = ['mysqlclient', 'iso8601', 'python-ldap', 'dirq', 'future']
+    # Python 2.x dependencies
+    if sys.version_info <= (3,):
```
tofu-rocketry, line 69:
```suggestion
    if sys.version_info < (3,):
```
Remember that you can't just flip the sign as the "3" is only inclusive on one side of the range.

test/test_norm_sum_record.py
```diff
@@ -15,6 +15,9 @@
 """Test cases for the NormalisedSummaryRecord based on SummaryRecord tests."""
 
 
+from builtins import zip
+from builtins import str
```
tofu-rocketry, line 19: Can we collapse these onto one line? (Like the `datetime` imports below.)

If that doesn't cause issues, can you do it in the other places that have more than one `builtins` import?

tofu-rocketry (changes requested):

setup.py
Sae126V: Resolved now.

test/test_norm_sum_record.py
Sae126V: Resolved now. Done

apel/db/loader/loader.py
```diff
@@ -17,10 +17,9 @@
 
 Module containing the Loader class.
 '''
+
 from __future__ import absolute_import
-from builtins import next
-from builtins import str
-from builtins import object
+from builtins import (next, str, object)
```
tofu-rocketry, line 22: Thanks for collating these, but the parentheses are unnecessary.

tofu-rocketry (changes requested):

apel/db/loader/loader.py
Sae126V: I followed the example given in the https://python-future.org/overview.html# @tofu-rocketry. Thus, followed the same convention for better readability.

tofu-rocketry: Change to `from future.builtins import next,`...

apel/db/apeldb.py
```diff
@@ -50,7 +52,7 @@ def __new__(cls, backend, host, port, username, pwd, db):
         except ImportError:
             logger.debug('Cannot import oracle backend')
 
-        if backend not in BACKENDS.keys():
+        if backend not in list(BACKENDS.keys()):
```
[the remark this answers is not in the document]
tofu-rocketry: @Sae126V, did you note this one?

Sae126V:
Hi @maxfischer2781. You might have seen the ticket https://github.com/apel/apel/issues/356; where we have captured the valuable points. We aim to respond on that soon. Hope it makes sense!

apel/db/apeldb.py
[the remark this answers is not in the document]
tofu-rocketry: Noted in #356

apel/db/loader/star_parser.py
```diff
@@ -117,8 +118,8 @@ def parseStarRecord(self, xml_storage_record):
 
         # Here we copy keys from functions.
         # We only want to change 'RecordId' to 'RecordIdentity'.
-        nodes = {}.fromkeys(map(lambda f: f == 'RecordId' and
-                                'RecordIdentity' or f, [S for S in functions]))
+        nodes = {}.fromkeys([f == 'RecordId' and
+                                'RecordIdentity' or f for f in [S for S in functions]])
```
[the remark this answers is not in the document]
tofu-rocketry: Noted in #356

apel/db/records/job.py
```diff
@@ -257,26 +259,26 @@ def get_ur(self, withhold_dns=False):
         service_level.appendChild(doc.createTextNode(str(self.get_field('ServiceLevel'))))
         ur.appendChild(service_level)
 
-        if self.get_field('MemoryReal') > 0:
+        if self.get_field('MemoryReal') is not None and int(self.get_field('MemoryReal')) > 0:
```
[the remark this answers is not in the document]
tofu-rocketry: Noted in #356

apel/ldap/query.py
```diff
@@ -115,15 +115,15 @@ def fetch_specint(site, host='lcg-bdii.egi.eu', port=2170):
...
+            except (KeyError, IndexError) as e:
                 log.error('Error during fetching Spec values: %s', e)
                 continue
 
             for fk in fks:
                 if fk.startswith(GLUE_CE_UNIQUE_ID):
                     name = fk.split('=')[1]
                     # do not overwrite values from first query
-                    if len(filter(lambda x: x[0] == name, values)) == 0:
+                    if len([x for x in values if x[0] == name]) == 0:
```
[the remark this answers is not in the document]
tofu-rocketry: Noted in #356

test/test_car_parser.py
tofu-rocketry: Noted in #356

tofu-rocketry (approved):
\* phew * I got through it all. Good job!

PR #367, opened 2024-05-10
Remove redundant Python 3 test from Travis and add caching to GH Action

tofu-rocketry:
We have working Python 3 tests in GitHub Actions.

This will save on build credits, which we currently have to request from Travis.


gregcorbett (approved):
LGTM

Issue #375, opened 2024-07-11
LDAP returns byte and not string with python3


vokac:
Apel client fails with exception when tries to use `startswith` with string argument on items returned from LDAP, because they have different data type (array of bytes). This prevents publishing accounting data in the EGI monitoring.

Is this still official tool for publishing EGI accounting data or we missed some announcement and it is necessary to use different tool with python3?

vokac:
btw: without too much thinking I "fixed" our client with
```patch
diff --git a/apel/ldap/query.py b/apel/ldap/query.py
index ba27ee2..f4c2569 100644
--- a/apel/ldap/query.py
+++ b/apel/ldap/query.py
@@ -79,13 +79,14 @@ def fetch_specint(site, host='lcg-bdii.egi.eu', port=2170):
     for entry in data:
 
         try:
-            ce_name = entry[1][GLUE_CE_UNIQUE_ID][0]
+            ce_name = entry[1][GLUE_CE_UNIQUE_ID][0].decode('utf-8')
             ce_capabilities=entry[1][GLUE_CE_CAPABILITY]
         except (KeyError, IndexError) as e:
             log.error('Error during fetching Spec values: %s', e)
             continue
 
         for capability in ce_capabilities:
+            capability = capability.decode('utf-8')
             si2k = parse_ce_capability(capability)
             if si2k is not None:
                 log.debug('Found value in first query: %s',
@@ -102,8 +103,8 @@ def fetch_specint(site, host='lcg-bdii.egi.eu', port=2170):
 
     for item in data:
         try:
-            cluster_name = item[1][GLUE_CHUNK_KEY][0].split('=')[1]
-            value = Decimal(item[1][GLUE_HOST_BENCHMARK][0])
+            cluster_name = item[1][GLUE_CHUNK_KEY][0].decode('utf-8').split('=')[1]
+            value = Decimal(item[1][GLUE_HOST_BENCHMARK][0].decode('utf-8'))
         except (KeyError, IndexError) as e:
             log.error('Error during fetching Spec values: %s', e)
             continue
@@ -120,10 +121,11 @@ def fetch_specint(site, host='lcg-bdii.egi.eu', port=2170):
                 continue
 
             for fk in fks:
+                fk = fk.decode('utf-8')
                 if fk.startswith(GLUE_CE_UNIQUE_ID):
                     name = fk.split('=')[1]
                     # do not overwrite values from first query
-                    if not any(x[0] == name for x in values):
[...]

PR #376, opened 2024-07-23
[GT-450] Ldap returns byte and not string with python3

Sae126V:
Resolves: https://github.com/apel/apel/issues/375

Issue: Apel client fails with exception when tries to use startswith with string argument on items returned from LDAP, because they have different data type (array of bytes)

### DIFF
--- a/apel/ldap/query.py
+++ b/apel/ldap/query.py
@@ -79,19 +79,21 @@ def fetch_specint(site, host='lcg-bdii.egi.eu', port=2170):
     for entry in data:
 
         try:
-            ce_name = entry[1][GLUE_CE_UNIQUE_ID][0]
+            ce_name = entry[1][GLUE_CE_UNIQUE_ID][0].decode('utf-8')
             ce_capabilities=entry[1][GLUE_CE_CAPABILITY]
         except (KeyError, IndexError) as e:
             log.error('Error during fetching Spec values: %s', e)
             continue
 
         for capability in ce_capabilities:
-            si2k = parse_ce_capability(capability)
+            capability_str = capability.decode('utf-8')
+            si2k = parse_ce_capability(capability_str)
+
             if si2k is not None:
                 log.debug('Found value in first query: %s',
-                          capability.split('=')[1])
+                          capability_str.split('=')[1])
                 values.append((ce_name,
-                           Decimal(capability.split('=')[1])))
+                           Decimal(capability_str.split('=')[1])))
 
     # not found? we have also second way to do this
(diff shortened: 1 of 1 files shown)


tofu-rocketry:
Could you note how this differs from @vokac's code and why please?

Sae126V:
> Could you note how this differs from @vokac's code and why please?

Hmm! Pretty much aligned with @vokac's that code itself. The Aim is we need to convert the bytes to strings.

We could tackle this bit differently if python-ldap < 4.0.0 versions. Since there is way for backward compatibility through `bytes_mode = True` for backwards compatible and `bytes_mode = False` for future compatible. Since `bytes_mode` is deprecated and ignored from python-ldap>4.0. Only way is to use .decode() to ensure to convert it into strings. Seems the file the ldap have might be in binary file.  So preferred .decode() to work with both versions.

tofu-rocketry:
> > Could you note how this differs from @vokac's code and why please?
> 
> Hmm! Pretty much aligned with @vokac's that code itself.

Ah, I see - the actual difference is that you're being clearer with the variable names once something is decoded.

> We could tackle this bit differently if python-ldap < 4.0.0 versions. Since there is way for backward compatibility through `bytes_mode = True` for backwards compatible and `bytes_mode = False` for future compatible in establishing the connection. Since `bytes_mode` is deprecated and ignored from python-ldap>4.0. So used .decode() to ensure it converts into strings. Seems the file the ldap reads might be in binary file itself. So preferred .decode() to work with both versions.

Yeah, decoding at the edges is the better way to do things. 👍🏻

apel/ldap/query.py
```diff
@@ -120,8 +122,10 @@ def fetch_specint(site, host='lcg-bdii.egi.eu', port=2170):
                 continue
 
             for fk in fks:
-                if fk.startswith(GLUE_CE_UNIQUE_ID):
-                    name = fk.split('=')[1]
+                ft_str = fk.decode('utf-8')
+
+                if ft_str.startswith(GLUE_CE_UNIQUE_ID):
+                    name = ft_str.split('=')[1]
```
tofu-rocketry, line 128:
```suggestion
                fk_str = fk.decode('utf-8')

                if fk_str.startswith(GLUE_CE_UNIQUE_ID):
                    name = fk_str.split('=')[1]
```

Looks like a typo in the variable name.

tofu-rocketry (commented):
Looks good. Just a typo that I'll sort out.

tofu-rocketry (approved):

PR #381, opened 2024-09-10
Deprecate separate logging config file

### LINKED ISSUE #293: Deprecate separate log config
`logging.cfg` causes a number of problems, not least that it doesn't seem to work and gives unexpected results if it's included without realising it. We should deprecate it and any related functionality (e.g. how it's handled in the executables).

Will-Cross1:
Remove separate logging file functionality as it's not generally used and causes issues when accidentally included.

much the same as apel/ssm#176 issue and apel/ssm#203 PR

Resolves #293
Resolves GT-541


bin/client.py
```diff
@@ -292,10 +292,16 @@ def main():
...
-                          default='/etc/apel/logging.cfg')
+                          help='DEPRECATED - location of logging config file (optional)',
+                          default=None)
 
     options, unused_args = opt_parser.parse_args()
+
+    # Deprecating functionality.
+    old_log_config_default_path = '/etc/apel/logging.cfg'
+    if (os.path.exists(old_log_config_default_path) or options.log_config is not None):
+        logging.warning('Separate logging config file option has been deprecated.')
```
tofu-rocketry, line 303: It would be good to move this after line 316, after the local logger is set up so that we can do `log.warning` so that the logging goes to the actual location that's configured. (Same for each file.)

The outer parentheses on line 302 are unnecessary.

tofu-rocketry (changes requested):

bin/client.py
```diff
@@ -292,10 +292,16 @@ def main():
...
     opt_parser.add_option('-l', '--log_config',
-                          help='location of logging config file (optional)',
-                          default='/etc/apel/logging.cfg')
+                          help='DEPRECATED - location of logging config file (optional)',
+                          default=None)
 
     options, unused_args = opt_parser.parse_args()
+
+    # Deprecating functionality.
+    old_log_config_default_path = '/etc/apel/logging.cfg'
```
tofu-rocketry, line 301: This "constant" only get used once in the next line, and the path is longer than the variable name! I'd say just fold that into the `if` statement.

bin/client.py
```diff
@@ -304,12 +310,9 @@ def main():
...
     try:
-        if os.path.exists(options.log_config):
-            logging.config.fileConfig(options.log_config)
-        else:
-            set_up_logging(ccp.get('logging', 'logfile'),
-                           ccp.get('logging', 'level'),
-                           ccp.getboolean('logging', 'console'))
+        set_up_logging(ccp.get('logging', 'logfile'),
+                        ccp.get('logging', 'level'),
+                        ccp.getboolean('logging', 'console'))
```
tofu-rocketry, line 315: Bit too much indentation.

tofu-rocketry (changes requested):

tofu-rocketry (changes requested):
Use `options.log_config` as the reference to the log config file path.

Remove the relevant argument from method signatures.

tofu-rocketry (changes requested):
Pretty much looks ok. Just one thing to do.

PR #399, opened 2024-11-20
Simplify argparse usage

### LINKED ISSUE #384: Simplify argparse usage?
Following #382 I'm wondering if it's possible to access the options more directly from the `NameSpace` object without using `vars` while still keeping the original workflow.

William-Brown5515:
Refactor argument parsing to use argparse Namespace object for structured access.

Resolves #384

### DIFF
--- a/bin/parser.py
+++ b/bin/parser.py
@@ -339,21 +339,24 @@ def main():
     ver = "APEL parser %s.%s.%s" % __version__
 
     default_conf_location = '/etc/apel/parser.cfg'
-    default_log_conf_location = '/etc/apel/parserlog.cfg'
     arg_parser = ArgumentParser(description=__doc__)
 
     arg_parser.add_argument('-c', '--config',
                             help='Location of config file',
                             default=default_conf_location)
     arg_parser.add_argument('-l', '--log_config',
-                            help='Location of logging config file',
-                            default=default_log_conf_location)
+                            help='DEPRECATED - Location of logging config file',
+                            default=None)
     arg_parser.add_argument('-v', '--version',
                             action='version',
                             version=ver)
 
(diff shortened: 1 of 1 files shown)


bin/dbloader.py
```diff
@@ -152,11 +151,11 @@ def run_as_daemon(loader, interval):
                             action='version',
                             version=ver)
 
-    # Using the vars function to output a dict-like view rather than Namespace object.
-    options = vars(arg_parser.parse_args())
+    # Parsing arguments into an argparse.Namespace object for structured access.
+    options: Namespace = arg_parser.parse_args()
```
tofu-rocketry, line 155: What's the benefit of using a type hint here? How necessary is it? It breaks the old Travis CI test (though that'll be going eventually).

William-Brown5515: There isn't - the comment above shows that the object is a Namespace object. I have removed the unnecessary bits.

bin/client.py
```diff
@@ -300,18 +300,18 @@ def main():
                             action='version',
```
tofu-rocketry, line 300: Need to remove the `(optional)` from the descriptions now it's deprecated.
Import ordering
Whitespace

Sae126V:
HI @tofu-rocketry , Are you planning to fix the order of the import in this PR ?

Otherwise, this PR looks good to me.

Sae126V (approved):
Ref: https://github.com/apel/apel/pull/399#issuecomment-2639736426

tofu-rocketry:
> HI @tofu-rocketry , Are you planning to fix the order of the import's in this PR ?
> 
> Otherwise, this PR looks good to me.

No. I generally prefer to keep that sort of change to a separate PR. This PR just removes an import.

PR #400, opened 2024-12-03
Update server-extra partitions for next few years

tofu-rocketry:


tofu-rocketry:
Did this get applied in the migration?

William-Brown5515 (approved):
LGTM

PR #402, opened 2024-12-06
Update to improve the naming convention for rpm builds

Sae126V:
See also: https://github.com/apel/ssm/pull/384


tofu-rocketry:
Going to decline this PR as I don't want to introduce any confusion with naming now that the build system is working reliably.

PR #403, opened 2024-12-10
Upgrade fields that match CloudSummary to BIGINT

### LINKED ISSUE #389: Cloud durations should be bigint
Durations in the cloud schema currently use `INT`, but some HPC usage of cloud records means that values are either truncated or rejected depending on DB setting for `STRICT_TRANS_TABLES`. Changing this would also bring it in line with the grid schema that already uses unsigned `BIGINT`.

Will need an upgrade schema to apply too.

tofu-rocketry:
Resolves #389 

There was an inconsistency in the CloudRecords table and view where the durations and some of the other fields used plain INT when the CloudSummaries table uses BIGINT. There have been records truncated or rejected due to this. This commit makes the two record types match in the DB schema (plus SuspendDuration for consistency with the other durations, even though it's not used in CloudSummaries).

DRAFT as I need to add an upgrade schema.


gregcorbett:
Looks good to me so far 😄

tofu-rocketry:
I don't think Codacy understands the bits where you change `DELIMITER` in SQL.

Sae126V (approved):
LGTM

PR #409, opened 2025-02-12
Replace assert with proper check

tofu-rocketry:
Asserts get skipped in compiled Python code, so shouldn't be relied on for normal checks within the code. Replace this with a check on the lengths and raise a ValueError if they don't match.


Sae126V (approved):
LGTM. 

We are raising the `ValueError` if the length is  NOT same than the AssertionError. Is this something we have to across codebase replacing assert with the proper check?

PR #416, opened 2025-03-27
Update codebase and the schema to change cpucount dbtype to decimal

Sae126V:
Resolves GT-809

TODO
scripts/update-2.3.0-x.x.x.sql -> Needs updating the script name.

### DIFF
--- a/apel/db/records/cloud.py
+++ b/apel/db/records/cloud.py
@@ -59,7 +59,7 @@ class CloudRecord(Record):
         self._int_fields = [ "SuspendDuration", "WallDuration", "CpuDuration",
                              "NetworkInbound", "NetworkOutbound", "PublicIPCount", "Memory", "Disk"]
 
-        self._float_fields = ['Benchmark', 'CpuCount']
+        self._float_fields = ['CpuCount', 'Benchmark']
         self._datetime_fields = ["RecordCreateTime", "StartTime", "EndTime"]
 
     def _check_fields(self):
--- a/apel/db/records/cloud.py
+++ b/apel/db/records/cloud.py
@@ -56,10 +56,10 @@ class CloudRecord(Record):
                                 "MeasurementMonth", "MeasurementYear"]
 
         # Fields which will have an integer stored in them
-        self._int_fields = [ "SuspendDuration", "WallDuration", "CpuDuration", "CpuCount",
+        self._int_fields = [ "SuspendDuration", "WallDuration", "CpuDuration",
                              "NetworkInbound", "NetworkOutbound", "PublicIPCount", "Memory", "Disk"]
 
-        self._float_fields = ['Benchmark']
+        self._float_fields = ['Benchmark', 'CpuCount']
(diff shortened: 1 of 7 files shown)


tofu-rocketry:
I've created a branch `intertwin-k8s-decimal-cpu` off `dev` for this to merge into, and I've changed the target branch for this PR.

apel/db/records/cloud.py
```diff
@@ -56,10 +56,10 @@ def __init__(self):
                                 "MeasurementMonth", "MeasurementYear"]
 
         # Fields which will have an integer stored in them
-        self._int_fields = [ "SuspendDuration", "WallDuration", "CpuDuration", "CpuCount",
+        self._int_fields = [ "SuspendDuration", "WallDuration", "CpuDuration",
                              "NetworkInbound", "NetworkOutbound", "PublicIPCount", "Memory", "Disk"]
 
-        self._float_fields = ['Benchmark']
+        self._float_fields = ['Benchmark', 'CpuCount']
```
tofu-rocketry, line 62: Can you swap the order to match the overall order in `msg_fields`?

apel/db/records/cloud_summary.py
```diff
@@ -53,8 +53,8 @@ def __init__(self):
 
         # Fields which will have an integer stored in them
         self._int_fields = ['Month', 'Year', 'WallDuration', 'CpuDuration',
-                            'CpuCount', 'NetworkInbound', 'NetworkOutbound',
+                            'NetworkInbound', 'NetworkOutbound',
                             'Memory', 'Disk', 'NumberOfVMs']
 
-        self._float_fields = ['Benchmark']
+        self._float_fields = ['Benchmark', 'CpuCount']
```
tofu-rocketry, line 59: Swap order.

schemas/cloud.sql
```diff
@@ -74,7 +74,7 @@ CREATE PROCEDURE ReplaceCloudRecord(
   startTime DATETIME, endTime DATETIME,
   suspendDuration BIGINT,
   wallDuration BIGINT, cpuDuration BIGINT,
-  cpuCount INT, networkType VARCHAR(255),  networkInbound BIGINT,
+  cpuCount DECIMAL(10,3), networkType VARCHAR(255),  networkInbound BIGINT,
```
tofu-rocketry, line 77: I think these should be `UNSIGNED`.

Am I correct in thinking that `(10,3)` will give us 7 leading digits and 3 trailing?

test/test_cloud_record.py
```diff
@@ -56,7 +56,7 @@ def setUp(self):
 SuspendDuration: NULL
 WallDuration: NULL
 CpuDuration: NULL
-CpuCount: 1
+CpuCount: 1.0
```
tofu-rocketry, line 59: Could some of these tests be changed so that they have a value after the decimal point?

Also, some of the data should still arrive with no decimal in it (as most cloud data is like that.

tofu-rocketry (changes requested):

tofu-rocketry (approved):

PR #421, opened 2025-06-30
Update unit test running to Ubuntu-22.04

tofu-rocketry:
This unfortunately means no Python 3.6 available, but the unit tests aren't running anyway as ubuntu-20.04 has been removed as a runner.

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -31,8 +31,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


Sae126V (approved):
LGTM!

Should we add 3.10 as well to the list, like SSM ?

tofu-rocketry:
> Should we add 3.10 as well to the list, like SSM ?

We don't need to worry about Ubuntu with APEL (just SSM), so don't need to add Python 3.10.

PR #425, opened 2025-07-10
Add config to recreate /var/run/apel folder on reboot

Sae126V:
Resolves [GT-987]

There will another PR in APEL-SSM codebase to resolve the issue with SSM rpms

[GT-987]: https://stfc.atlassian.net/browse/GT-987?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/apel-build.sh
+++ b/scripts/apel-build.sh
@@ -111,6 +111,16 @@ mkdir -p "$TEMP_DIR_FOR_SERVER/var/run/apel"
 # apel-lib
 mkdir -p "$TEMP_DIR_FOR_LIB/$PYTHON_ROOT_DIR/$LIB_EXTENSION/apel/"
 
+# Ensure /var/run/apel is created at every reboot by systemd-tmpfiles
+# https://www.man7.org/linux/man-pages/man5/tmpfiles.d.5.html
+mkdir -p "$TEMP_DIR_FOR_SERVER/etc/tmpfiles.d"
+
+# Create a tmpfiles.d *.conf file using `cat` to manage the creation of /var/run/apel directory
+# https://www.uptimia.com/learn/linux-cat-command-useful-examples
+cat > "$TEMP_DIR_FOR_SERVER/etc/tmpfiles.d/run-apel.conf" <<EOF
+d /var/run/apel 0755 root root -
+EOF
+
 APEL_DIR=apel-$VERSION-$ITERATION
 
 # Download and extract source
@@ -240,6 +250,7 @@ FPM_SERVER_PACKAGE="-n apel-server \
     --config-files /etc/apel/loader.cfg \
     --config-files /etc/apel/db.cfg \
     --config-files /etc/apel/auth.cfg \
+    --config-files /etc/tmpfiles.d/run-apel.conf \
     --description \"APEL server package\" "
 
 FPM_LIB_PACKAGE="-n apel-lib \
--- a/scripts/apel-build.sh
+++ b/scripts/apel-build.sh
@@ -113,11 +113,11 @@ mkdir -p "$TEMP_DIR_FOR_LIB/$PYTHON_ROOT_DIR/$LIB_EXTENSION/apel/"
 
 # Ensure /var/run/apel is created at every reboot by systemd-tmpfiles
 # https://www.man7.org/linux/man-pages/man5/tmpfiles.d.5.html
-mkdir -p "$TEMP_DIR_FOR_SERVER/etc/tmpfiles.d"
+mkdir -p "$TEMP_DIR_FOR_SERVER/usr/lib/tmpfiles.d"
 
 # Create a tmpfiles.d *.conf file using `cat` to manage the creation of /var/run/apel directory
 # https://www.uptimia.com/learn/linux-cat-command-useful-examples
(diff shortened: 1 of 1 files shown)


scripts/apel-build.sh
```diff
@@ -111,6 +111,12 @@ mkdir -p "$TEMP_DIR_FOR_SERVER/var/run/apel"
 # apel-lib
 mkdir -p "$TEMP_DIR_FOR_LIB/$PYTHON_ROOT_DIR/$LIB_EXTENSION/apel/"
 
+# Ensure systemd's tmpfiles creates the /var/run/apel folder on every reboot
+mkdir -p "$TEMP_DIR_FOR_SERVER/etc/tmpfiles.d"
+cat > "$TEMP_DIR_FOR_SERVER/etc/tmpfiles.d/run-apel.conf" <<EOF
```
gregcorbett: It took me a while to parse what was happening here, it's a use of `cat` I hadn't come across before. Maybe an additional comment explaining what it is doing would be useful.

gregcorbett (dismissed):
Looks good, an explanation of why this works and a link to some documentation would be useful in the commit message.

scripts/apel-build.sh
Sae126V: Added a link and the enhanced the commit message. Please let me know if it make sense.

tofu-rocketry:
Will there be any conflict with the change also being applied to SSM as per #apel/ssm/pull/405 ?

Sae126V:
> Will there be any conflict with the change also being applied to SSM as per #[apel/ssm/pull/405](https://github.com/apel/ssm/pull/405) ?

It will NOT complain as APEL and SSM ship their own `*.conf` file in `tmpfiles.d`. So, `systemd-tmpfiles` will gracefully handle if the folder exists at the reboot.

scripts/apel-build.sh
```diff
@@ -111,6 +111,16 @@ mkdir -p "$TEMP_DIR_FOR_SERVER/var/run/apel"
 # apel-lib
 mkdir -p "$TEMP_DIR_FOR_LIB/$PYTHON_ROOT_DIR/$LIB_EXTENSION/apel/"
 
+# Ensure /var/run/apel is created at every reboot by systemd-tmpfiles
+# https://www.man7.org/linux/man-pages/man5/tmpfiles.d.5.html
+mkdir -p "$TEMP_DIR_FOR_SERVER/etc/tmpfiles.d"
```
tofu-rocketry: Install to /usr/lib/tmpfiles.d

scripts/apel-build.sh
```diff
@@ -111,6 +111,16 @@ mkdir -p "$TEMP_DIR_FOR_SERVER/var/run/apel"
...
+
+# Create a tmpfiles.d *.conf file using `cat` to manage the creation of /var/run/apel directory
+# https://www.uptimia.com/learn/linux-cat-command-useful-examples
+cat > "$TEMP_DIR_FOR_SERVER/etc/tmpfiles.d/run-apel.conf" <<EOF
```
tofu-rocketry: Name should be as per the rpm - apel-server

tofu-rocketry (changes requested):

scripts/apel-build.sh
Sae126V: Done. I installed the configuration file in /etc/tmpfiles because it overrides any other file with the same name. Similarly, I noticed that /etc/logrotate.d is located within the /etc directory, which is why I chose to place it there as well -  to ensure it takes precedence over any existing file with the same name.

scripts/apel-build.sh
Sae126V: Done

scripts/apel-build.sh
tofu-rocketry: Note that even what you might think of as quite low level packages use `/usr/lib/tmpfiles.d`. `/etc/tmpfiles.d` is for local admin usage to override the package installed settings as per the [man page](https://www.man7.org/linux/man-pages/man5/tmpfiles.d.5.html).

e.g.
```
# ls /usr/lib/tmpfiles.d/
README                         rpcbind.conf
audit.conf                     screen.conf
cloud-init.conf                selinux-policy.conf
cockpit-ws.conf                setroubleshoot.conf
criu.conf                      setup.conf
cryptsetup.conf                static-nodes-permissions.conf
dnf.conf                       sudo.conf
etc.conf                       systemd-nologin.conf
home.conf                      systemd-nspawn.conf
journal-nocow.conf             systemd-pstore.conf
legacy.conf                    systemd-tmp.conf
libselinux.conf                systemd.conf
man-db.conf                    tmp.conf
mdadm.conf                     tpm2-tss-fapi.conf
pam.conf                       udisks2.conf
podman.conf                    var.conf
provision.conf                 x11.conf
rootfiles.conf

# ls /etc/tmpfiles.d/
sos.conf

```

Sae126V: I am using `/usr/lib/tmpfiles.d`  now 👍. Make sense to me, Adrian.

tofu-rocketry (approved):

PR #427, opened 2025-08-14
Update codebase and the schema to change cpu count type to decimal

tofu-rocketry:
Will need the schema update script finalising in terms of version numbers and rebasing once 2.3.1 is done.

To check:
- [x] That `DECIMAL(10,3)` is big enough for existing `CpuCount` values or if we need (13,3) for safety.
-> Biggest value in DB is 128 CPUs. ✅ 

This code was tested as part of the interTwin project and ran successfully there. Requires #441 to enable the option to unload decimal CPU counts as decimal or integer (for EGI Portal compatibility).

### DIFF
--- a/scripts/update-2.3.0-x.x.x.sql
+++ b/scripts/update-2.3.0-x.x.x.sql
@@ -1,0 +1,117 @@
+-- This script contains a SQL block that can update
+-- APEL version 2.3.0 databases of the following types to x.x.x:
+--  - Cloud Accounting Database
+
+-- UPDATE SCRIPT FOR CLOUD SCHEMA
+
+-- This section will:
+-- - Update CpuCount DBType in CloudRecords to DECIMAL(10,3) UNSIGNED from INT.
+-- - Update the ReplaceCloudRecord procedure to match this change.
+-- - Update CpuCount DBType in CloudSummaries to DECIMAL(10,3) UNSIGNED from INT.
+-- - Update the ReplaceCloudSummaryRecord procedure to match this change.
+
+-- CloudRecords
+ALTER TABLE CloudRecords MODIFY CpuCount DECIMAL(10,3) UNSIGNED;
+
+
+DROP PROCEDURE IF EXISTS ReplaceCloudRecord;
+DELIMITER //
+CREATE PROCEDURE ReplaceCloudRecord(
+  recordCreateTime DATETIME,VMUUID VARCHAR(255), site VARCHAR(255), cloudComputeService VARCHAR(255),
+  machineName VARCHAR(255),
+  localUserId VARCHAR(255),
+  localGroupId VARCHAR(255), globalUserName VARCHAR(255),
+  fqan VARCHAR(255), vo VARCHAR(255),
+  voGroup VARCHAR(255), voRole VARCHAR(255), status VARCHAR(255),
+  startTime DATETIME, endTime DATETIME,
+  suspendDuration BIGINT,
+  wallDuration BIGINT, cpuDuration BIGINT,
+  cpuCount DECIMAL(10,3) UNSIGNED, networkType VARCHAR(255),  networkInbound BIGINT,
+  networkOutbound BIGINT, publicIPCount INT, memory BIGINT,
+  disk BIGINT, benchmarkType VARCHAR(50), benchmark DECIMAL(10,3), storageRecordId VARCHAR(255),
+  imageId VARCHAR(255), cloudType VARCHAR(255),
+  publisherDN VARCHAR(255))
+
+BEGIN
+    DECLARE measurementTimeCalculated DATETIME;
+    DECLARE recordCreateTimeNotNull DATETIME;
+    DECLARE endTimeNotNull DATETIME;
+
+    IF(status='completed') THEN
+        -- in this case, the recordCreateTime and measurementTime could
+        -- be wildly different as the VM has ended.
+
+        -- if we werent supplied a record create time
+        -- for a completed VM we have decided to use the end time
+        -- We have to check the EndTime is not null because we can't guarantee this as we
+        -- previously loaded completed messages with no EndTime.
+        SET endTimeNotNull = IFNULL(endTime, '0000-00-00 00:00:00');
+        SET recordCreateTimeNotNull = IF(recordCreateTime IS NULL or recordCreateTime='0000-00-00 00:00:00', endTimeNotNull, recordCreateTime);
+
+        -- Use the end time as the measurement time
+        SET measurementTimeCalculated = endTimeNotNull;
+
+    ELSE
+        -- In the case of a running VM, the measurement time will
+        -- equal the record create time
+        IF(recordCreateTime IS NULL or recordCreateTime='0000-00-00 00:00:00') THEN
+            -- Calculate the time of measurement so we can use it later to determine which
+            -- accounting period this incoming record belongs too.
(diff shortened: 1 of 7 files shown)


test/test_cloud_record.py
```diff
@@ -56,7 +56,7 @@ def setUp(self):
 SuspendDuration: NULL
 WallDuration: NULL
 CpuDuration: NULL
-CpuCount: 1
+CpuCount: 1.0000
```
rowan04:
```suggestion
CpuCount: 1.000
```
is this supposed to have 4 zeroes? The others only have three digits after a decimal point.

scripts/update-2.3.0-x.x.x.sql
```diff
@@ -0,0 +1,117 @@
...
+-- - Update CpuCount DBType in CloudRecords to DECIMAL(10,3) UNSIGNED from INT.
+-- - Update the ReplaceCloudRecord procedure to match this change.
+-- - Update CpuCount DBType in CloudSummaries to DECIMAL(10,3) UNSIGNED from INT.
+-- - Update the ReplaceCloudSummaryRecord procedure to match this change.
+
+-- CloudRecords
+ALTER TABLE CloudRecords MODIFY CpuCount DECIMAL(10,3) UNSIGNED;
+
+
+DROP PROCEDURE IF EXISTS ReplaceCloudRecord;
```
garaimanoj, line 17: Are we going to call these stored procedures only once? How are we going to call?

scripts/update-2.3.0-x.x.x.sql
```diff
@@ -0,0 +1,117 @@
...
+    END IF;
+
+    REPLACE INTO CloudRecords(RecordCreateTime, VMUUID, SiteID, CloudComputeServiceID, MachineName,
+        LocalUserId, LocalGroupId, GlobalUserNameID, FQAN, VOID, VOGroupID,
+        VORoleID, Status, StartTime, EndTime, MeasurementTime, MeasurementMonth,
+        MeasurementYear, SuspendDuration, WallDuration, CpuDuration, CpuCount,
+        NetworkType, NetworkInbound, NetworkOutbound, PublicIPCount, Memory, Disk,
+        BenchmarkType, Benchmark, StorageRecordId, ImageId, CloudType, PublisherDNID)
+      VALUES (
+        recordCreateTimeNotNull, VMUUID, SiteLookup(site), CloudComputeServiceLookup(cloudComputeService), machineName,
```
garaimanoj, line 81: In a stored procedure, assign the lookup results to local variables first might improve the performance. 
```
SET siteID_val = SiteLookup(site);
SET serviceID_val = CloudComputeServiceLookup(cloudComputeService);
```
But to make sure we many test the performance if needed by
```
SET @start = NOW(6);
CALL your_procedure_name();
SET @end = NOW(6);

SELECT TIMEDIFF(@end, @start) AS total_execution_time;
```

garaimanoj (approved):

scripts/update-2.3.0-x.x.x.sql
tofu-rocketry: This is the update script so it overwrites the old versions of procedures. This script is only intended to be run once during the upgrade process.

scripts/update-2.3.0-x.x.x.sql
```diff
@@ -0,0 +1,117 @@
...
+
+    ELSE
+        -- In the case of a running VM, the measurement time will
+        -- equal the record create time
+        IF(recordCreateTime IS NULL or recordCreateTime='0000-00-00 00:00:00') THEN
+            -- Calculate the time of measurement so we can use it later to determine which
+            -- accounting period this incoming record belongs too.
+            SET measurementTimeCalculated = TIMESTAMPADD(SECOND, (IFNULL(suspendDuration, 0) + IFNULL(wallDuration, 0)), startTime);
+            -- We recieve and currently accept messages without a start time
+            -- which causes the mesaurementTimeCalculated to be NULL
```
rowan04:
```suggestion
            -- accounting period this incoming record belongs to.
            SET measurementTimeCalculated = TIMESTAMPADD(SECOND, (IFNULL(suspendDuration, 0) + IFNULL(wallDuration, 0)), startTime);
            -- We receive and currently accept messages without a start time
            -- which causes the measurementTimeCalculated to be NULL
```

scripts/update-2.3.0-x.x.x.sql
```diff
@@ -0,0 +1,117 @@
...
+        LocalUserId, LocalGroupId, GlobalUserNameID, FQAN, VOID, VOGroupID,
+        VORoleID, Status, StartTime, EndTime, MeasurementTime, MeasurementMonth,
+        MeasurementYear, SuspendDuration, WallDuration, CpuDuration, CpuCount,
+        NetworkType, NetworkInbound, NetworkOutbound, PublicIPCount, Memory, Disk,
+        BenchmarkType, Benchmark, StorageRecordId, ImageId, CloudType, PublisherDNID)
+      VALUES (
+        recordCreateTimeNotNull, VMUUID, SiteLookup(site), CloudComputeServiceLookup(cloudComputeService), machineName,
+        localUserId, localGroupId, DNLookup(globalUserName), fqan, VOLookup(vo), VOGroupLookup(voGroup),
+        VORoleLookup(voRole), status, startTime, endTime, measurementTimeCalculated, Month(measurementTimeCalculated), Year(measurementTimeCalculated),
+        suspendDuration, wallDuration, cpuDuration, cpuCount,
```
rowan04:
```suggestion
        VORoleLookup(voRole), status, startTime, endTime, measurementTimeCalculated, Month(measurementTimeCalculated),
        Year(measurementTimeCalculated), suspendDuration, wallDuration, cpuDuration, cpuCount,
```
this shortens quite a long line and also matches how the lines are broken up in the `REPLACE INTO` above better? Allows for better readability.

scripts/update-2.3.0-x.x.x.sql
```diff
@@ -0,0 +1,117 @@
...
+  publisherDN VARCHAR(255))
+BEGIN
+    REPLACE INTO CloudSummaries(SiteID, CloudComputeServiceID, Month, Year, GlobalUserNameID, VOID, VOGroupID, VORoleID, Status, CloudType, ImageId, EarliestStartTime, LatestStartTime,
+        WallDuration, CpuDuration, CpuCount, NetworkInbound, NetworkOutbound, Memory, Disk, BenchmarkType, Benchmark, NumberOfVMs,  PublisherDNID)
+      VALUES (
+        SiteLookup(site), CloudComputeServiceLookup(cloudComputeService), month, year, DNLookup(globalUserName), VOLookup(vo),
+        VOGroupLookup(voGroup), VORoleLookup(voRole), status, cloudType, imageId, earliestStartTime, latestStartTime,
+        wallDuration, cpuDuration, cpuCount, networkInbound, networkOutbound, memory,
+        disk, benchmarkType, benchmark, numberOfVMs, DNLookup(publisherDN)
+        );
```
rowan04:
```suggestion
      );
```

scripts/update-2.3.0-x.x.x.sql
tofu-rocketry: Noted here for future reference: https://github.com/apel/apel/issues/446

For now, this it out of scope of this PR as the update script mainly recreates the code already existing in the `cloud.sql` schema (with tweaks as per the diff in this PR).

PR #440, opened 2026-01-28
Update codeql-analysis.yml

tofu-rocketry:
This change adds Actions to the things checked by CodeQL and stops forks from running checks.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,4 +1,4 @@
-name: "CodeQL Advanced"
+name: "CodeQL"
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
rowan04, line 51: the new file for monitoring added in https://github.com/apel/monitoring/pull/79/changes has these extra lines for with category - should these be added here as well?

Image

tofu-rocketry: I think that's bit of an unnecessary bit as the category is already automatically set based on the matrix. Might remove it from the other YAML.

tofu-rocketry: Removed from https://github.com/apel/monitoring/pull/79

rowan04 (approved):
lgtm!

PR #441, opened 2026-02-02
Set config value to upload optional decimal cpu count

garaimanoj:
Resolves GT-1297

Manual testing:
- Update config file at `/etc/apel/cloud/cloud-unloader.cfg`
```
table_name=VCloudSummaries
decimal_cpu_count = true
```
- Update latest code for `unloader.py` at `/usr/lib/python3.9/site-packages/apel/db`
- Update `apeldbunloader` at `/usr/bin`
- Set up database schema and table
- Populate `/etc/apel/cloud/cloud-db.cfg` with appropriate value
- Run `/usr/bin/apeldbunloader -d /etc/apel/cloud/cloud-db.cfg -c /etc/apel/cloud/cloud-unloader.cfg`
- Test the result at `/var/spool/apel/cloud/outgoing`

### DIFF
--- a/apel/db/records/cloud.py
+++ b/apel/db/records/cloud.py
@@ -133,6 +133,19 @@ class CloudRecord(Record):
             raise InvalidRecordException("Cannot parse an integer from StartTime or EndTime.")
 
 
+    '''
+    Move a field from one type category to another.
+
+    This updates the internal type lists by removing `field_name` from the
+    list associated with `from_type` and appending it to the list associated
+    with `to_type`. The method only performs the change if the field is
+    currently present in the source type list.
+    Notes
+    -----
+    - No action is taken if from_type is not a valid key in the internal
+      type mapping or if the field is not present in the source list.
+    - This method assumes that to_type is a valid key in the type map.
+    '''
     def change_field_type(self, field_name, from_type, to_type):
         type_map = {
             'float': self._float_fields,
--- a/apel/db/records/cloud.py
+++ b/apel/db/records/cloud.py
@@ -131,3 +131,14 @@ class CloudRecord(Record):
 
         except (ValueError, OverflowError, OSError):
             raise InvalidRecordException("Cannot parse an integer from StartTime or EndTime.")
+
+
+    def change_field_type(self, field_name, from_type, to_type):
+        type_map = {
+            'float': self._float_fields,
+            'int': self._int_fields
+        }
+
+        if from_type in type_map and field_name in type_map[from_type]:
+            type_map[from_type].remove(field_name)
+            type_map[to_type].append(field_name)
--- a/apel/db/unloader.py
+++ b/apel/db/unloader.py
@@ -240,7 +240,7 @@ class DbUnloader(object):
         for batch in self._db.get_records(record_type, table_name, query=query, records_per_message=self.records_per_message):
             if record_type == CloudRecord and not self._decimal_cpu_count:
                 for row in batch:
-                    row._record_content['CpuCount'] = self._to_int_min1(row._record_content.get('CpuCount'))
+                    row.set_field('CpuCount', self._to_int_min1(row.get_field('CpuCount')))
 
             records += len(batch)
             if ur:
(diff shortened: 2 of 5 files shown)


apel/db/unloader.py
```diff
@@ -245,6 +250,37 @@
             msgs += 1
 
         return msgs, records
+    
+    def _to_int_min1(v):
```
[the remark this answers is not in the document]
garaimanoj: self added

apel/db/unloader.py
```diff
@@ -246,6 +251,39 @@ def _write_messages(self, record_type, table_name, query, ur):
...
         return msgs, records
 
+    def _to_int_min1(self, v):
+        '''
+        Convert a value to an integer with a special rule:
+        - If the numeric value is less than 1 and greater than 0 (0 < v < 1), return 1.
+        - For all other numeric values, return int(v).
+        - Return None if the value cannot be interpreted as a number.
+
+        Examples:
```
tofu-rocketry: Rather than provide a list of examples, this would be better as a unit test of the function.

bin/dbunloader.py
```diff
@@ -158,6 +158,11 @@ def _bounded_records_per_message(config_object, logger):
     except ConfigParser.NoOptionError:
         withhold_dns = False
 
+    try:
+        decimal_cpu_count     = cp.getboolean('unloader', 'decimal_cpu_count')
```
tofu-rocketry: Remove the extra whitespace. It doesn't help it align to anything anyway.

conf/unloader.cfg
```diff
@@ -55,6 +55,9 @@ interval = latest
 gap_start = 2014-01-01
 gap_end = 2014-01-31
 
+# Option to either unload decimal or integer CPU count for Cloud records from repo to portal
```
tofu-rocketry:
```suggestion
# For cloud records only, unload decimal CPU counts if available in the DB (if true) or coerce to integers (if false).
```
To make it a bit clearer what it does

tofu-rocketry (changes requested):
A reminder to please put the "why" in commit messages in future.

tofu-rocketry (changes requested):
A reminder to please put the "why" in commit messages in future.

bin/dbunloader.py
garaimanoj: Done

conf/unloader.cfg
garaimanoj: Done

apel/db/unloader.py
garaimanoj: Done

tofu-rocketry:
Just noting that the call to `DbUnloader` in [`bin/client.py`](https://github.com/apel/apel/blob/09f27779890c7dbec5b60c1726739c4a12b3ad2a/bin/client.py#L240) hasn't been changed, but the new argument is optional and we don't need it for the client side.

tofu-rocketry (approved):
Looks good. Gonna hold off merging until I have time to do both the storage related PRs together.

garaimanoj:
Output after changing the type of field

image

apel/db/records/cloud.py
```diff
@@ -131,3 +131,14 @@ def _check_start_end_times(self):
 
         except (ValueError, OverflowError, OSError):
             raise InvalidRecordException("Cannot parse an integer from StartTime or EndTime.")
+
+
+    def change_field_type(self, field_name, from_type, to_type):
```
tofu-rocketry, line 149: Could do with a docstring.

tofu-rocketry (changes requested):
I think the fix is OK, but it does feel a little bit over-complicated for what it does - a good docstring would help alleviate that though.

apel/db/records/cloud.py
garaimanoj: Done

tofu-rocketry (approved):
Looks good. Thanks.

PR #442, opened 2026-02-17
Add InfrastructureDescription to appropriate places in client and ser…

Sae126V:
Resolves [GT-1300](https://stfc.atlassian.net/jira/software/c/projects/GT/boards/70?assignee=6406fcdf0d9b61193c26c89d&selectedIssue=GT-1300)

[GT-1300]: https://stfc.atlassian.net/browse/GT-1300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apel/db/unloader.py
```diff
@@ -292,9 +293,23 @@ def _write_apel(self, records):
...
+        # Build the set of fields to exclude based on configuration.
+        exclude_fields = set()
+        
+        if record_type in (
+            SummaryRecord,
+            SummaryRecord04,
+            NormalisedSummaryRecord,
+            NormalisedSummaryRecord04
+        ):
+            if not self.include_infrastructure_description:
```
tofu-rocketry:
```suggestion
        ) and not self.include_infrastructure_description:
```

This would be a bit simpler using an `and` wouldn't it?

apel/db/records/record.py
```diff
@@ -274,7 +274,14 @@ def get_msg(self, withhold_dns=False):
         None.  In this case, no line is included in the message unless
         it is a mandatory field.  If the field is mandatory, an
         exception is raised.
+
+        If exclude_fields is provided, any field in that set will be omitted
```
tofu-rocketry: What type should `exclude_fields` be provided as?

test/test_norm_sum_record.py
```diff
@@ -191,19 +193,22 @@ def _get_record_tuples(self):
...
+                  '/atlas', 'Role=production', 'some.host.org', 'local',
+                  'APEL-CREAM-HTCONDOR', None,
                   None, None, datetime.utcfromtimestamp(1270083600),
                   datetime.utcfromtimestamp(1271725200), 234256, 244435, 234256,
                   244435, 100)
 
         tuple3 = ('RAL-LCG2', 12, 2010, '/C=whatever/D=someDN', 'atlas',
-                  '/atlas', 'Role=production', 'some.host.org', 'local', None, 1, 1,
+                  '/atlas', 'Role=production', 'some.host.org', 'local',
+                  'None', None, 1, 1,
```
tofu-rocketry, line 213: It comes out as the string "None" if it's not defined? Is that right?

apel/db/records/record.py
Sae126V: This is to do with adding the fields itself, Adrian. For instance, like InfrastructureDescription field we can say either include or exclude the field from the outgoing message when unloading.

test/test_norm_sum_record.py
Sae126V: yep. To my understanding, The string values are getting 'None' (not None).

apel/db/records/record.py
tofu-rocketry: I meant that the docstring isn't clear on the Python type of the argument.

tofu-rocketry (commented):
That looks good now. We'll move onto some thorough testing next Sprint.

apel/db/records/normalised_summary.py
```diff
@@ -196,6 +198,8 @@ def get_ur(self):
 
         infra = doc.createElement('aur:Infrastructure')
         infra.setAttribute('urf:type', self.get_field('Infrastructure'))
+        if self.get_field('InfrastructureDescription') is not None:
+            infra.setAttribute('urf:description', self.get_field('InfrastructureDescription'))
```
tofu-rocketry: This shouldn't be touching the AUR format.

bin/client.py
```diff
@@ -239,6 +239,9 @@ def run_client(ccp):
 
         unloader = DbUnloader(db, unload_dir, include_vos, exclude_vos,
                               local_jobs, withhold_dns, dict_records)
+        # Preserve existing client behaviour: Ensures to include `infrastructureDescription` in outgoing messages.
+        unloader.include_infrastructure_description = True
+
```
tofu-rocketry: Why is this included? This PR doesn't actually make changes to the client schema, and isn't the existing behaviour to _not_ include?

scripts/update-2.3.0-2.3.1.sql
```diff
@@ -0,0 +1,359 @@
+-- ============================================================================
+-- APEL UPDATE SCRIPT FOR SERVER SCHEMA
+-- APEL version 2.3.0 databases of the following types to 2.3.1:
```
tofu-rocketry: Don't assume the version numbers when creating update scripts. That's a task for the release branch.

apel/db/records/normalised_summary.py
Sae126V: Done...Removed this

bin/client.py
Sae126V: > Why is this included?

To Preserve the existing behavior for the client as-is. To ensure client code works as it is working before introducing this change.

> and isn't the existing behaviour to not include?

No. It was including in the client code flow.

scripts/update-2.3.0-2.3.1.sql
Sae126V: Cool. I think I have missed this one.

bin/client.py
tofu-rocketry: Remove these lines.

apel/db/backends/mysql.py
```diff
@@ -73,10 +73,10 @@ class ApelMysqlDb(object):
...
-              SummaryRecord: "CALL ReplaceSummary(%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)",
-              SummaryRecord04: "CALL ReplaceSummary(%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)",
-              NormalisedSummaryRecord: "CALL ReplaceNormalisedSummary(%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)",
-              NormalisedSummaryRecord04: "CALL ReplaceNormalisedSummary(%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)",
+              SummaryRecord: "CALL ReplaceSummary(%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)",
+              SummaryRecord04: "CALL ReplaceSummary(%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)",
+              NormalisedSummaryRecord: "CALL ReplaceNormalisedSummary(%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)",
+              NormalisedSummaryRecord04: "CALL ReplaceNormalisedSummary(%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)",
```
tofu-rocketry, line 79: Am I correct in thinking that there procedures aren't used by the client at all? Just checking we're not going to break the client inadvertently by changing the call signature.

apel/db/records/record.py
```diff
@@ -274,7 +274,14 @@ def get_msg(self, withhold_dns=False):
         None.  In this case, no line is included in the message unless
         it is a mandatory field.  If the field is mandatory, an
         exception is raised.
+
+        If exclude_fields is provided(as a set of strings), any field name in that set
+        will be omitted from the output message. This is used to conditionally exclude
+        optional fields like InfrastructureDescription based on configuration.
         '''
+        if exclude_fields is None:
+            exclude_fields = set()
```
tofu-rocketry, line 283: Why a set, and not just a tuple?

bin/client.py
```diff
@@ -239,6 +239,7 @@ def run_client(ccp):
 
         unloader = DbUnloader(db, unload_dir, include_vos, exclude_vos,
                               local_jobs, withhold_dns, dict_records)
+
```
tofu-rocketry: Still got an errant whitespace change.

tofu-rocketry (changes requested):
Think that looks ok. Just one minor change and a couple of clarifying questions.

apel/db/backends/mysql.py
Sae126V: > procedures aren't used by the client at all?

No. Not at this moment. `VSuperSummaries` is in both

apel/db/records/record.py
Sae126V: Any preference you have, Adrian. I have used for storing the unique fields.

tofu-rocketry: Fair enough.

tofu-rocketry (approved):

PR #445, opened 2026-03-09
Convert cursorclass named arguments to positional in cursor calls

### LINKED ISSUE #443: Convert "cursorclass" named arguments to positional in cursor calls
To improve compatibility with different Python DB packages, "cursorlcass" named arguments should be converted to plain positional arguments as some packages use the named argument `cursorclass` while others just use `cursor`. Both that I've seen have the cursor class as the only possible argument, so removing the name should make the APEL code more widely compatible.

Occurrences (but double check for others):
https://github.com/apel/apel/blob/09f27779890c7dbec5b60c1726739c4a12b3ad2a/scripts/migrate_apel.py#L129
https://github.com/apel/apel/blob/09f27779890c7dbec5b60c1726739c4a12b3ad2a/apel/db/backends/mysql.py#L167
https://github.com/apel/apel/blob/09f27779890c7dbec5b60c1726739c4a12b3ad2a/apel/db/backends/mysql.py#L230

Example result:
```python
self.db.cursor(MySQLdb.cursors.DictCursor)
```

rowan04:
- as some python db packages have differing arg naming conventions (e.g: cursorclass and cursor)
- but both seen so far have it as the only positional argument

Resolves #443
Resolves GT-1372

### DIFF
--- a/apel/db/backends/mysql.py
+++ b/apel/db/backends/mysql.py
@@ -164,7 +164,7 @@ class ApelMysqlDb(object):
             # 'MySQL server has gone' exception
             self._mysql_reconnect()
 
-            c = self.db.cursor(cursorclass=MySQLdb.cursors.DictCursor)
+            c = self.db.cursor(MySQLdb.cursors.DictCursor)
 
             for record in record_list:
                 values = record.get_db_tuple(source)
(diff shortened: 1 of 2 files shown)


tofu-rocketry (approved):
Nice. Good commit message too.

PR #464, opened 2026-06-15
Extend coercion to integer for decimal CPU counts to Cloud Summaries

### LINKED ISSUE #463: Extend coercion to integer for decimal CPU counts to Cloud Summaries
In #411 we missed that the change is really needed for `CloudSummaryRecord`, rather than just the `CloudRecord`.

The method `change_field_type` should be moved up to the `Record` class so it's accessible from all sub-classes.

The record classes the change applies to should be extended to `CloudSummaryRecord` here https://github.com/apel/apel/blob/53dbaaa4d8c47b1a27ff88fd3e560ff617076393/apel/db/unloader.py#L242

garaimanoj:
Resolves #463 

Manual Testing Steps

1. Update `/etc/apel/cloud/cloud-unloader.cfg`:
    - Set decimal_cpu_count to true, then repeat with false.
    - Set table_name to VCloudRecords, then repeat with VCloudSummaries.
2. Run `/usr/bin/apeldbunloader -d /etc/apel/cloud/cloud-db.cfg -c /etc/apel/cloud/cloud-unloader.cfg`
3. After each configuration change, verify the output generated in:
`/var/spool/apel/cloud/outgoing`

### DIFF
--- a/apel/db/records/cloud.py
+++ b/apel/db/records/cloud.py
@@ -131,27 +131,3 @@ class CloudRecord(Record):
 
         except (ValueError, OverflowError, OSError):
             raise InvalidRecordException("Cannot parse an integer from StartTime or EndTime.")
-
-
-    '''
-    Move a field from one type category to another.
-
-    This updates the internal type lists by removing `field_name` from the
-    list associated with `from_type` and appending it to the list associated
-    with `to_type`. The method only performs the change if the field is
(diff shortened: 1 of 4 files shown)


tofu-rocketry (approved):
Good work. I've just chucked in a basic unit test for the `change_field_type` method.

PR #466, opened 2026-06-26
Add new method to get APEL version

garaimanoj:
Add new method to get APEL version to generate infra description for summaries

image

### DIFF
--- a/apel/parsers/parser.py
+++ b/apel/parsers/parser.py
@@ -68,7 +68,7 @@ class Parser(object):
 
         return False
     
-    def get_client_version():
+    def get_client_version(self):
         '''
         Method to get APEL client version
         '''
--- a/apel/parsers/parser.py
+++ b/apel/parsers/parser.py
@@ -17,6 +17,7 @@
 '''
 
 from future.builtins import object
+from apel import __version__
 
 import logging
 
@@ -66,3 +67,9 @@ class Parser(object):
             pass
 
         return False
+    
+    def get_client_version():
+        '''
+        Method to get APEL client version
+        '''
+        return 'APEL/%s.%s.%s' % __version__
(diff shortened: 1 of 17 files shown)


apel/parsers/parser.py
```diff
@@ -66,3 +67,9 @@ def recognize(self, line):
             pass
 
         return False
+
+    def get_client_version(self):
+        '''
+        Method to get APEL client version
+        '''
+        return f"APEL/{'.'.join(map(str, __version__))}"
```
tofu-rocketry: This isn't particularly readable.

APEL generally uses %-formatting for the version number, `'%i.%i.%i' % __version__`, or a more readable version of this would be something like `"APEL/{0}.{1}.{2}".format(*__version__)`

schemas/client.sql
```diff
@@ -444,6 +444,7 @@ CREATE TABLE SuperSummaries (
   VORoleID INT NOT NULL,                -- Foreign key
```
tofu-rocketry, line 444: We'll need an update script to go with these schema changes.

apel/parsers/parser.py
garaimanoj: Done

schemas/client.sql
garaimanoj: Script added. Against which database do we need to run this script?

tofu-rocketry (approved):
Looks good.

schemas/client.sql
tofu-rocketry: As discussed, the client version of APEL is only used with the client schema, so we don't need to specify other than saying that it applies to the client db.

tofu-rocketry:
Thinking I'll merge this into a 3.0.0 release as we may want to use the major version increment to signal to users of the client that there's a backwards incompatible change (i.e. schema change).

PR #469, opened 2026-07-15
Drop SuperSummaries table

### LINKED ISSUE #458: Actually deprecate the `SuperSummaries` table in the server schema
The `SuperSummaries` table was deprecated a long time ago. We should actually remove it.

https://github.com/apel/apel/blob/0580763fee45150b6b6b6726bcd68de4ec305413/schemas/server.sql#L201-L206

(The view `VSuperSummaries` now takes its data from `HybridSuperSummaries` and doesn't need touching.)

garaimanoj:
Resolves #458

### DIFF
--- a/apel/db/backends/mysql.py
+++ b/apel/db/backends/mysql.py
@@ -208,7 +208,7 @@ class ApelMysqlDb(object):
 
     def get_sync_records(self, query=None, records_per_message=1000):
         """
-        Get sync records from the SuperSummaries table. Filter by the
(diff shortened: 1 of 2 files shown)


tofu-rocketry (changes requested):
Changes look good, but the commit message needs the context of "why".

(You can squash the commits.)

tofu-rocketry (approved):
Hold off merge until 2.6.0 is released.

PR #470, opened 2026-07-15
Replace assertions with ValueError

### LINKED ISSUE #412: Replace `assert`s in SGEParser
Similar to #409 , we should replace these with a check that raises some kind of parsing error.

garaimanoj:
Resolved: https://github.com/apel/apel/issues/412
Replace assertions with ValueError to ensures validation is always performed, even when Python assertions are disabled

### DIFF
--- a/apel/parsers/sge.py
+++ b/apel/parsers/sge.py
@@ -156,10 +156,12 @@ class SGEParser(Parser):
         for key in mapping:
             data[key] = mapping[key](values)
 
-        assert data['CpuDuration'] >= 0, 'Negative CpuDuration value'
-        assert data['WallDuration'] >= 0, 'Negative WallDuration value'
-        assert data['StopTime'] > 0, 'Zero epoch time for field StopTime'
-
+        if data['CpuDuration'] < 0:
+            raise ValueError("Negative CpuDuration value: %s" % data['CpuDuration'])
+        if data['WallDuration'] < 0:
+            raise ValueError("Negative WallDuration value: %s" % data['WallDuration'])
+        if data['StopTime'] <= 0:
+            raise ValueError("Zero epoch time for field StopTime")
 
         record.set_all(data)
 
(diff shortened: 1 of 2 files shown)


apel/parsers/sge.py
```diff
@@ -155,10 +155,12 @@ def parse(self, line):
...
-        assert data['CpuDuration'] >= 0, 'Negative CpuDuration value'
-        assert data['WallDuration'] >= 0, 'Negative WallDuration value'
-        assert data['StopTime'] > 0, 'Zero epoch time for field StopTime'
-
+        if data['CpuDuration'] < 0:
+            raise ValueError(f"Negative CpuDuration value: {data['CpuDuration']}")
+        if data['WallDuration'] < 0:
+            raise ValueError(f"Negative WallDuration value: {data['WallDuration']}")
+        if data['StopTime'] <= 0:
+            raise ValueError('Zero epoch time for field StopTime')
```
tofu-rocketry:
```suggestion
            raise ValueError("Zero epoch time for field StopTime")
```

apel/parsers/sge.py
```diff
@@ -155,10 +155,12 @@ def parse(self, line):
         for key in mapping:
             data[key] = mapping[key](values)
 
-        assert data['CpuDuration'] >= 0, 'Negative CpuDuration value'
-        assert data['WallDuration'] >= 0, 'Negative WallDuration value'
-        assert data['StopTime'] > 0, 'Zero epoch time for field StopTime'
-
+        if data['CpuDuration'] < 0:
+            raise ValueError(f"Negative CpuDuration value: {data['CpuDuration']}")
```
tofu-rocketry: We don't use f-strings anywhere else in the codebase, so can you swap this and the other one to %-formatting, please?

tofu-rocketry (changes requested):

apel/parsers/sge.py
garaimanoj: Done

apel/parsers/sge.py
garaimanoj: Done

tofu-rocketry (approved):
Hold off merge until 2.6.0 is released.

PR #474, opened 2026-07-16
Remove Python 2 compatibility bits

### LINKED ISSUE #383: Remove Python 2 compatability bits
See similar work in SSM, particularly in https://github.com/apel/ssm/milestone/24?closed=1

Also:

- [ ] Remove Python 2 build bits
- [ ] Set shebangs to Python 3

garaimanoj:
Resolved: https://github.com/apel/apel/issues/383

I have added two more changes apart from the tasks in the ticket

- Remove Python version check as it is unnecessary
- Use issubset() for subset assertions in tests


tofu-rocketry (changes requested):
Double check that it is that case, but if we've removed all usage of the `future` module, it can be removed from `requirements.txt` the spec file and anywhere else it's included as a dependency.

garaimanoj:
> Double check that it is that case, but if we've removed all usage of the `future` module, it can be removed from `requirements.txt` the spec file and anywhere else it's included as a dependency.

Done

tofu-rocketry (approved):
Good work. I've just added a couple of commits. One to tidy up the import ordering, and the other to remove an unused Python 2 function.

apel/db/loader/loader.py
```diff
@@ -18,9 +18,6 @@
 Module containing the Loader class.
 '''
 
-from __future__ import absolute_import
-from future.builtins import next, str
-
 import logging
 import os
 import sys
```
Sae126V: This is being NOT used. Since we removed the sys.version_info >= (3,). Can this be removed?

tofu-rocketry: Good point. I'll sort that.

PR #477, opened 2026-08-17
Fix accelerator records and summaries schema

Sae126V:
Resolves [GT-1699]
(testing noted in Jira)

[GT-1699]: https://stfc.atlassian.net/browse/GT-1699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/schemas/accelerator.sql
+++ b/schemas/accelerator.sql
@@ -70,7 +70,7 @@ CREATE TABLE AcceleratorRecords (
   MeasurementMonth INT NOT NULL,
   MeasurementYear INT NOT NULL,
 
-  AssociatedRecordType VARCHAR(255) NOT NULL,
+  AssociatedRecordType VARCHAR(50) NOT NULL,
   AssociatedRecord VARCHAR(255) NOT NULL,
 
   GlobalUserName VARCHAR(255),
@@ -82,7 +82,7 @@ CREATE TABLE AcceleratorRecords (
   AvailableDuration INT,
   BenchmarkType VARCHAR(255),
   Benchmark DECIMAL(10,3),
-  Type VARCHAR(255) NOT NULL,
+  Type VARCHAR(50) NOT NULL,
   Model VARCHAR(255) NOT NULL,
   PublisherDNID INT NOT NULL,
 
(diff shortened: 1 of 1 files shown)


schemas/accelerator.sql
```diff
@@ -345,7 +345,10 @@ DROP PROCEDURE IF EXISTS GetModelSummaries;
 DELIMITER //
 CREATE PROCEDURE GetModelSummaries()
 BEGIN
-    REPLACE INTO AcceleratorModelSummaries
+    REPLACE INTO AcceleratorModelSummaries (
```
tofu-rocketry, line 348: I saw the commit message, but just wondering, do you know where the difference in the number of columns came from?

tofu-rocketry (commented):
Thanks for looking at this quickly. Also, good commit messages!

Could you note any testing done in the PR's opening comment, please?

schemas/accelerator.sql
Sae126V: > do you know where the difference in the number of columns came from?

I am not able to understand the question - But, It is the MariaDB complaining that columns should match the values.

Sae126V:
> Could you note any testing done in the PR's opening comment, please?

I have added in the JIRA ticket itself - Do you want me to replicate that here ?

tofu-rocketry:
Ah, right, no need then. I've just added a note to the opening comment to signpost to it.

schemas/accelerator.sql
tofu-rocketry: Ah, right, I've got it - it's because of the `UpdateTime` column that sets itself. There's 15 columns in the `REPLACE` and `SELECT`, but the table has 16 because of the additional `UpdateTime` column.

tofu-rocketry: Could you add that last sentence to the commit message for some extra detail?
> There's 15 columns in the REPLACE and SELECT, but the table has 16 because of the additional UpdateTime column.

Sae126V: Done.

tofu-rocketry (approved):
Good work.

garaimanoj (approved):
LGTM

